PR #14, opened 2017-12-04
Removed Docker related files 

iler:
Remove Docker related files as the official repositories are not using Docker to test the provider implementations. Let's follow the way the testing is done in official repositories.


vtorhonen:
Docker was not used for testing but for running & deploying the provider through a job scheduler. It's no longer needed since Terraform provides their own official containers.

PR #17, opened 2018-03-07
Adjusted the installation instructions and example instructions

### LINKED ISSUE #16: Documentation is incomplete even for testing
The documentation in both the root directory (https://github.com/vtorhonen/terraform-provider-upcloud/blob/master/README.md) and examples diretory (https://github.com/vtorhonen/terraform-provider-upcloud/blob/master/examples/README.md) is incomplete and conflicting even for testing the provider. If one follows the instructions for installing in the provider in the root directory and tries to run `terraform init` it ends up with the following error:

```
Initializing provider plugins...
- Checking for available provider plugins on https://releases.hashicorp.com...

Provider "upcloud" not available for installation.

A provider named "upcloud" could not be found in the official repository.

This may result from mistyping the provider name, or the given provider may
be a third-party provider that cannot be installed automatically.

In the latter case, the plugin must be installed manually by locating and
downloading a suitable distribution package and placing the plugin's executable
file in the following directory:
    terraform.d/plugins/darwin_amd64

Terraform detects necessary plugins by inspecting the configuration and state.
To view the provider versions requested by each module, run
"terraform providers".
```

aleksijohansson:
Simplified the installation instructions of the go code and added a step to symlink the provider into a folder where Terraform will find it when running `terraform init`.

This fixes #16.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,18 +22,18 @@ Requirements
 Building The Provider
 ---------------------
 
-Clone repository to: `$GOPATH/src/github.com/terraform-providers/terraform-provider-upcloud`
+Get and install the provider:
 
 ```sh
-$ mkdir -p $GOPATH/src/github.com/terraform-providers; cd $GOPATH/src/github.com/terraform-providers
-$ git clone git@github.com:vtorhonen/terraform-provider-upcloud.git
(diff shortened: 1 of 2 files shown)


vtorhonen:
Thanks!

PR #20, opened 2018-04-10
Updated server and other schemas

meafmira:


vtorhonen:
Thanks, does this implement the whole UpCloud API? I'll need to have a closer look at this and do some testing this week. Could you resolve the merge conflict in the meantime?

Issue #32, opened 2018-07-11
Firewall rule `position` required but not caught in `plan`


Johannestegner:
When creating a new firewall rule through the API the `position` argument is optional (in the documentation that is), the same seems to apply for the `terraform plan` command but not for the `terraform apply` command.  

The position field is used when querying for the rules upon removal or updates, which makes it understandable that they require it, but it would be great if the `plan` command threw errors too!

iler:
The fix is now merged. Thanks 👍

PR #33, opened 2018-07-11
Changed position in firewall rule to Required instead of Optional.

Johannestegner:
As mentioned in #32, the `position` argument in firewall rules is required, it is though, in the schema, set to optional.
This PR changes it to Required.


iler:
@Johannestegner thanks for this! Merging it.

PR #34, opened 2018-07-12
Fix for tags.

### LINKED ISSUE #31: Tagging feature doesn't work
I have the following tf file, and am running into issues while using the tagging feature. Used latest version of provider 

```
$ terraform --version
Terraform v0.11.7
+ provider.upcloud (unversioned)
```

**resource.tf**
```

resource "upcloud_server" "s2s" {

  zone = "nl-pbp1"
  hostname = "s2s.example.com"

  cpu = "2"
  mem = "4096"

  ipv6 = false

  storage_devices = [
    {
      action = "clone"
      size = 10
      address = "virtio:0"
      storage = "01e3dbaf-c236-448a-bf03-5b43ad626140"
      tier = "maxiops"
    }
  ]

}

resource "upcloud_tag" "core" {
  description = "Core functionality"
  name = "core"
  servers = {
    server = [
     "${upcloud_server.s2s.id}"
   ]
  }
  depends_on = [
    "upcloud_server.s2s",
  ]
}
```

I get the following error when I try to do apply or see the plan

```bash

$ terraform plan

Error: upcloud_tag.core: servers: unexpected map value type: map[string]*schema.Schema{"server":*schema.Schema{Type:5, Optional:false, Required:true, DiffSuppressFunc:(schema.SchemaDiffSuppressFunc)(nil), Default:interface {}(nil), DefaultFunc:(schema.SchemaDefaultFunc)(nil), Description:"", InputDefault:"", Computed:false, ForceNew:false, StateFunc:(schema.SchemaStateFunc)(nil), Elem:4, MaxItems:0, MinItems:0, PromoteSingle:false, Set:(schema.SchemaSetFunc)(nil), ComputedWhen:[]string(nil), ConflictsWith:[]string(nil), Deprecated:"", Removed:"", ValidateFunc:(schema.SchemaValidateFunc)(nil), Sensitive:false}}

```

Johannestegner:
Due to the definition of `Servers` in the `Upcloud.Tag` struct in `upcloud-go-api` the API expects a `Servers` parameter with a string list. This is not totally compatible with the API documentation which expects a map with a key named `server` and a value which consist of a `[]string` of IDs.  
This PR makes the terraform API use a list directly on the `servers` key to match the `upcloud-go-api` structure.

To define a set of tags the following is sufficient with this PR:

```hcl
resource "upcloud_tag" "My-tag" {
  name        = "TagName"
  description = "TagDescription"
  servers     = [
     "server-id-1",
     "server-id-2"
  ]
}
```

This should fix #31, although it does not match the http API docs.

### DIFF
--- a/upcloud/resource_upcloud_tag.go
+++ b/upcloud/resource_upcloud_tag.go
@@ -28,14 +28,10 @@ func resourceUpCloudTag() *schema.Resource {
 				ForceNew: true,
 			},
 			"servers": {
-				Type:     schema.TypeMap,
-				Optional: true,
-				Elem: map[string]*schema.Schema{
-					"server": {
-						Type:     schema.TypeList,
-						Required: true,
-						Elem:     schema.TypeString,
-					},
+				Type:     schema.TypeList,
+				Required: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
 				},
 			},
 		},
@@ -58,8 +54,12 @@ func resourceUpCloudTagCreate(d *schema.ResourceData, meta interface{}) error {
 		createTagRequest.Description = description.(string)
 	}
 	if servers, ok := d.GetOk("servers"); ok {
-		servers := servers.(map[string]interface{})
-		createTagRequest.Servers = servers["server"].([]string)
+		servers := servers.([]interface{})
+		serversList := make([]string, len(servers))
+		for i := range serversList {
+			serversList[i] = servers[i].(string)
+		}
+		createTagRequest.Servers = serversList
 	}
 
 	tag, err := client.CreateTag(createTagRequest)


cancerballs:
Thank you!

cancerballs:
With some further testing a bug was discovered.

When starting with a situation where one server is tagged:

`resource "upcloud_tag" "My-tag" {
  name        = "TagName1"
  description = "TagDescription"
  servers     = [
     "${upcloud_server.test.id}"
  ]
}
`
And then adding more servers to be tagged as follows

`resource "upcloud_tag" "My-tag" {
  name        = "TagName1"
  description = "TagDescription"
  servers     = [
     "${upcloud_server.test.id}"
  ]
}
`

Terraform crashes and outputs the error

`* upcloud_tag.My-tag: unexpected EOF
2018/07/25 12:29:54 [ERROR] root: eval: *terraform.EvalSequence, err: 1 error(s) occurred:`
 
I'm not going to revert this fix (at least for now) because it makes tagging possible but this bug ought to be fixed and instead of a new issue it should be fixed in this branch where the bug originated.

Johannestegner:
Did you create a new `resource` for the second tag (with a new name and tag) or did you define two resources with the same name? Just so I don't start debugging and it ends up as a terraform limitation, hehe.

cancerballs:
As you can see in the example code I first created a resource with one server, then after running apply I added a second server to the same resource.

So going from 1 to 2 servers in the same resource crashed terraform.

Johannestegner:
Will take a look at it right away :)

Johannestegner:
Yes sir, did a PR a while back! #40 :)

PR #54, opened 2018-10-16
Fixed terraform acceptance tests

nikohaap:
* Fixed all Terraform acceptance tests related on storage, server(instance) and firewall
* Optimized acceptance tests storage sizes
* Added some validators for instance parameters in the schema

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,9 +1,9 @@
-package main // import "github.com/UpCloudLtd/terraform-provider-upcloud"
+package main
 
 import (
 	"github.com/hashicorp/terraform/plugin"
 	"github.com/hashicorp/terraform/terraform"
-	"github.com/UpCloudLtd/terraform-provider-upcloud/upcloud"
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -1,9 +1,9 @@
-package main // import "github.com/UpCloudLtd/terraform-provider-upcloud"
+package main
 
 import (
 	"github.com/hashicorp/terraform/plugin"
 	"github.com/hashicorp/terraform/terraform"
-	"github.com/UpCloudLtd/terraform-provider-upcloud/upcloud"
+	"github.com/nikohaap/terraform-provider-upcloud/upcloud"
```
cancerballs: Can you clarify the purpose of this change?

nikohaap: It was jut because I was using the forked package instead of the UpCloudLtd. 
It should be the UpCloudLtd version instead of the forked (nikohaap) one :)

PR #59, opened 2018-11-05
Higher timeout for creating servers

derkgort:
I've set the timeout higher to accommodate the longer creation times in Chicago and Singapore.
Also, When creating it is probably better to have it take longer than to retry it a few times..


aelbarkani:
would be possible to merge this PR ? we regularly have issues with Upcloud servers not being provisioned within the 5 limit

LordChocoBlizzBallz:
Thanks for the PR.

PR #75, opened 2020-07-28
Reivew documentation and examples

trentrosenbaum:
The focus of this PR is to confirm the existing documentation and examples or the UpCloud Terraform provider.  The provider is a community driven provider and document/example need to flow for a user of the provider.  This will be the aim for the updates.

Alterations
- Adjusted content covering requirements, (versions) and command to install/build provider.


trentrosenbaum:
The destination was not correct and referenced the UpCloud instead of OpenCredo

PR #85, opened 2020-10-26
feat: add user-agent to header

PopoSensei:
First try at adding header in provider.


aakso (changes requested):

tohalla:
Played around with this, 

- now sets the header through the client
- script for syncing the version in globals to the latest tag

still need to figure out a place for running the sync script before shipping

Issue #86, opened 2020-10-27
source_ip_filtering always stuck to true


malafoss:
Upcloud terraform provider does not let source_ip_filtering to be set to false.

### Steps to reproduce:

Apply the following terraform file:
```
terraform {
  required_providers {
    upcloud = {
      source = "UpCloudLtd/upcloud"
      version = "1.0.0"
    }
  }
}

provider "upcloud" {
}

resource "upcloud_network" "network" {
  name = "my private network"
  zone = "fi-hel1"

  ip_network {
    address = "10.201.0.0/24"
    dhcp = false
    dhcp_default_route = false
    family = "IPv4"
    gateway = "10.201.0.1"
  }
}

resource "upcloud_server" "server" {
  zone = "fi-hel1"
  hostname = "server"
  plan = "1xCPU-1GB"

  storage_devices {
    size = 10
    action = "clone"
    tier = "maxiops"
    title = "boot disk"
    storage = "Debian GNU/Linux 10 (Buster)"
  }

  network_interface {
    type = "private"
    network = upcloud_network.network.id
    source_ip_filtering = false
  }
}

output "source_ip_filtering" {
  value = upcloud_server.server.network_interface[0].source_ip_filtering
}
```

### Expected behaviour:

Should output `source_ip_filtering = false`.

### Actual behaviour:

Incorrectly outputs `source_ip_filtering = true`, which also causes recreation of new instance if another `terraform apply` is made subsequently. Possibly calls to `POST /1.3/server/{uuid}/networking/interface` are not made at all, which also prevents use of static ip addresses.

isarpila:
Hi,

Thanks for the feedback. We were able to reproduce this problem and are working to fix it.

BR,

-Ilkka

PR #92, opened 2020-11-04
Fix falses not being set correctly

tohalla:
getOk requires non-zero value. `False` is the zero value for booleans.

For ref, see #86 

source_ip_filtering

### DIFF
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -148,14 +148,14 @@ func resourceUpCloudServer() *schema.Resource {
 							Description: "`true` if source IP should be filtered.",
 							ForceNew:    true,
 							Optional:    true,
-							Computed:    true,
+							Default:     true,
 						},
 						"bootable": {
 							Type:        schema.TypeBool,
 							Description: "`true` if this interface should be used for network booting.",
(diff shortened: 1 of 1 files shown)


upcloud/resource_upcloud_server.go
```diff
@@ -712,11 +711,11 @@ func buildNetworkOpts(d *schema.ResourceData, meta interface{}) ([]request.Creat
 			Type: d.Get(keyRoot + "type").(string),
 		}
 
-		if v, ok := d.GetOk(keyRoot + "source_ip_filtering"); ok {
+		if v, ok := d.GetOkExists(keyRoot + "source_ip_filtering"); ok {
```
PopoSensei, line 714: >// Deprecated: usage is discouraged due to undefined behaviors and may be
>// removed in a future version of the SDK

This is when I looked into what this does.
Hmmm. Though they have changed the comment now to read. When I looked at the github repo.

>// This should only be used if absolutely required/needed.

There are more in the comment, but I guess this could be used here.

tohalla: updated to use .Get instead. Also replaced config in the schema to have default values

PopoSensei:
LGTM 🌮

PR #94, opened 2020-11-10
Fix/server resource inconsistencies

tohalla:
- update on hostname should not ForceNew
- drop unsupported field from server resource (template)
- fix incorrect descriptions on fields


tohalla:
https://github.com/UpCloudLtd/terraform-provider-upcloud/pull/96 for allowing updating of the hostname

LordChocoBlizzBallz:
Looks good.

PR #95, opened 2020-11-10
fix: should not be able to set plan with cpu / mem specified

tohalla:
the values have to match and becomes tedious to upkeep (also not setting the plan will result to a flexible server being created)


PopoSensei:
LGTM 🌮  👍

PR #98, opened 2020-11-13
fix: setting firewall rules

tohalla:
firewall rules cannot be modified when the server is in `maintenance` state


LordChocoBlizzBallz:
Looks good.

PR #99, opened 2020-11-26
upgrade upcloud-go-api

tohalla:


vendor/github.com/UpCloudLtd/upcloud-go-api/internal/globals.go
```diff
@@ -0,0 +1,3 @@
+package globals
+
+var Version = "3.0.0"
```
PopoSensei, line 3: Should we release a new version of the go-api and then upgrade to that version, since there are at least the user-agent stuff that we are interested at.

tohalla: Pushed the latest stuff, alright to merge? Need to get this in to wrap up with the storage stuff. Currently updated by hash (not pointing to any release).

PR #104, opened 2020-12-18
Refactor/separate storages from server resource

tohalla:
- Type fixes
- Fix sync issues (when state changed outside terraform)
- Changes in storage devices (or partially in template) will not trigger recreation of the server resource)
- Only the template will be recreated when server resource is recreated
- The template is managed within server, when deleting a server, also the template storage will be destroyed. (previously used to delete all the other storages as well)
- The "normal" storage devices will be managed as separate resources (need to remove the declaration to remove the storage. Removing storage from within the server block just detaches it). Can be attached to servers with:
    ```tf
    resource "upcloud_server" "server_resource_name" {
        zone     = "fi-hel1"
        ...
        storage_devices {
            address = "scsi"
            storage = upcloud_storage.storage_name.id # can also set to fixed uuid
            type = "disk"
        }
        storage_devices {
            storage = upcloud_storage.another_storage_name.id # can also set to fixed uuid
        }
    }

    resource "upload_storage" "storage_name" {
        # backup_rule { ... }
        clone: "target storage uuid"
        # import { ... }
        size = 25
        tier = "maxiops"
        title = "Storage device name"
        zone = "fi-hel1"
    }
    resource "upload_storage" "another_storage_name" {
        ...
    }
    ```
    - **NOTE**: did not rename the clone field, might want to either rename it or template.storage for consistency.

- Templates (introduced in this mr) differ from storages a bit. The templates are tied to server deployment, as they are setup when deploying a server. One can declare template followingly:

    ```tf
    resource "upcloud_server" "server_resource_name" {
        zone     = "fi-hel1"
        ...
        template {
          address: "scsi"
          # backup_rule: { ... }
          storage: "uuid for the public template"
          title: "Storage device name"
        }
    }
    ```
    - Dropped the `tier` from here, as for the plans hdd storages would be billed separately

## Other notes

- Did not address the code organisation here, something that should be discussed later on
- The acceptance tests need some love (will be addressed in a separate pr)

### DIFF
--- a/upcloud/resource_upcloud_floating_ip_address_test.go
+++ b/upcloud/resource_upcloud_floating_ip_address_test.go
@@ -189,11 +189,9 @@ func testUpcloudFloatingIPAddressCreateWithServerConfig(serverNames []string, as
   			hostname = "mydebian.example.com"
   			plan     = "1xCPU-2GB"
 
-  			storage_devices {
-    			action = "create"
-    			size   = 10
-    			tier   = "maxiops"
-  			}
+        template {
+          storage = "01000000-0000-4000-8000-000020050100"
+        }
 
   			network_interface {
     			type = "public"
--- a/upcloud/resource_upcloud_floating_ip_address_test.go
+++ b/upcloud/resource_upcloud_floating_ip_address_test.go
@@ -189,9 +189,9 @@ func testUpcloudFloatingIPAddressCreateWithServerConfig(serverNames []string, as
   			hostname = "mydebian.example.com"
   			plan     = "1xCPU-2GB"
 
-        template {
-          storage = "01000000-0000-4000-8000-000020050100"
-        }
+  			template {
+  				storage = "01000000-0000-4000-8000-000020050100"
+  			}
 
   			network_interface {
     			type = "public"
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -166,7 +166,8 @@ func resourceUpCloudServer() *schema.Resource {
 			},
 			"storage_devices": {
 				Description: "A list of storage devices associated with the server",
-				Type:        schema.TypeSet,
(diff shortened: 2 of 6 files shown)


tohalla:
Still plenty work to do with the tests

tohalla:
Also contains changes from #99

upcloud/resource_upcloud_floating_ip_address_test.go
```diff
@@ -189,11 +189,9 @@ func testUpcloudFloatingIPAddressCreateWithServerConfig(serverNames []string, as
   			hostname = "mydebian.example.com"
   			plan     = "1xCPU-2GB"
 
-  			storage_devices {
-    			action = "create"
-    			size   = 10
-    			tier   = "maxiops"
-  			}
+        template {
+          storage = "01000000-0000-4000-8000-000020050100"
```
ajmyyra, line 193: Not that big of a problem, but there's different indentation here and in some other places where this PR has touched.

upcloud/resource_upcloud_server.go
```diff
@@ -516,155 +562,69 @@ func buildServerOpts(d *schema.ResourceData, meta interface{}) (*request.CreateS
...
-			for _, s := range l.Storages {
-				if s.Title == source {
-					source = s.UUID
-					break
-				}
-			}
+			serverStorageDevice.Storage = source
+		}
+		r.StorageDevices = append(r.StorageDevices, serverStorageDevice)
+		// TODO: handle backup_rule
```
ajmyyra, line 602: Was this forgotten or just documented for future work?

ajmyyra (approved):
As the old Finnish saying goes, kunnon hyvä! 💪

upcloud/resource_upcloud_server.go
tohalla: was forgotten. good catch!

PR #107, opened 2020-12-27
Expose the ability to explicitly enable or disable metadata service.

joliver:
Updated the Terraform provider the provide the ability to enable or disable the per-instance metadata service.


ajmyyra (approved):
Hi @joliver and thank you for the PR! I'm confused why this was previously forgotten, but at least it's now there and will find itself to the next version we're publishing soonish. :)

Would recommend merging this immediately, unless it's conflicting with something @tohalla is working on.

joliver:
Excellent! Thanks for approving this. Any ideas when it might be merged and then published? We'd like to start using UpCloud but we're waiting on this PR before we start using the service.

PR #110, opened 2021-01-19
Fix/acceptance tests

tohalla:


.github/workflows/acceptance_tests.yml
```diff
@@ -0,0 +1,23 @@
+name: UpCloud Terraform provider acceptance tests
+
+on:
+  push:
+    branches:
+      - master
```
PopoSensei, line 6: Why not on PR:s?

tohalla: the acceptance tests take almost three hours to run 😁

tohalla: should probably look into that at some point as well

PopoSensei: Yep. I've been wondering about that too.

PR #111, opened 2021-01-19
feature: WithRetry utility fn

tohalla:
occasionally we want to re-attempt running a function in case it fails.

E.g. dispatching storage detach request at the same time causes the server go to a maintenance state (storages cannot be detached when server in maintenance state) for a short while. We don't want the provider to throw the failure user (while only one storage detached) when he attempts to fire multiple storage detach request at one go, so we'd use the helper fn to reattempt detaching storages.

### DIFF
--- a/upcloud/utils.go
+++ b/upcloud/utils.go
@@ -61,13 +61,13 @@ func min(x, y int) int {
 	return y
 }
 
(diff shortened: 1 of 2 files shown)


tohalla:
Will be merged to master along 2.0.0

PR #113, opened 2021-01-19
Update examples

Darep:
This PR:

- adds extra options, optimal defaults and comments to 01_server.tf example
- adds floating IP address example
- adds remote commend execution example

### DIFF
--- a/examples/01_server.tf
+++ b/examples/01_server.tf
@@ -1,36 +1,40 @@
 provider "upcloud" {
-  # You need to set UpCloud credentials in shell environment variable
-  # using .bashrc, .zshrc or similar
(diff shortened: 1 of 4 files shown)


tohalla:
Looking good! Would love if we additionally had one example with minimal config required

Darep:
> Looking good! Would love if we additionally had one example with minimal config required

Thanks! Let's see if i can add that as well :)

Darep:
Added example of minimal config now too!

tohalla:
btw, was thinking if the numeric prefixes should be removed from the filenames, would be more clear that these are individual examples rather than parts of a shared config

PR #128, opened 2021-01-29
Improve gogetcookie.sh

szepeviktor:


scripts/gogetcookie.sh
```diff
@@ -1,10 +1,13 @@
 #!/bin/bash
+#
+# Create googlesource.com cookie
+#
+# From https://github.com/hashicorp/terraform/blob/master/scripts/gogetcookie.sh
 
-touch ~/.gitcookies
-chmod 0600 ~/.gitcookies
+install -m 0600 /dev/null ~/.gitcookies
```
szepeviktor, line 7: This change empties the gitcookies file instead of appending to it.

tohalla:
Thanks for the PR. 

Closing this, as having the script around is not necessary anymore (will drop it altogether)

szepeviktor:
# 😲

PR #129, opened 2021-02-01
chore: drop unnecessary scripts

tohalla:


tohalla:
acceptance tests fail here due to previous residues (and not having sweepers up)

Darep (approved):
LGTM!

PR #132, opened 2021-03-25
feat(server): add host field

tohalla:
add support for explicitly selecting host (for private cloud customers)

(also fixes server not starting when updating storages, was referring to storage id previously)


kaminek:
LGTM, need to rebase

kaminek (approved):

PR #136, opened 2021-04-27
tests: fix some test after modules update

kaminek:


LordChocoBlizzBallz (approved):
Looks good. Can be merged into master even tho tests fail because those are related to object storage tests and will be fixed shortly.

PR #138, opened 2021-05-04
Bugfix refactor objstore tests

LordChocoBlizzBallz:


upcloud/resource_upcloud_objectstorage_test.go
```diff
@@ -349,16 +349,20 @@ func verifyObjectStorageExists(accessKey, secretKey, name string) resource.TestC
 
 func verifyObjectStorageDoesNotExist(accessKey, secretKey, name string) resource.TestCheckFunc {
 	return func(state *terraform.State) error {
-		time.Sleep(time.Second * 3)
-		exists, err := doesObjectStorageExists(state, accessKey, secretKey)
```
kaminek, line 353: can you delete the trailing function and all its dependancies

kaminek (commented):
Is it the the only test that fails, iirc there were like 3 failing

upcloud/resource_upcloud_objectstorage_test.go
```diff
@@ -349,16 +349,20 @@ func verifyObjectStorageExists(accessKey, secretKey, name string) resource.TestC
...
-			if err.Error() == "could not find resources" {
-				return nil
+
+		for _, rs := range state.RootModule().Resources {
+			if rs.Type != "upcloud_storage" {
+				continue
+			}
+
+			client := testAccProvider.Meta().(*service.Service)
+			_, err := client.GetObjectStorageDetails(&request.GetObjectStorageDetailsRequest{
```
moitias, line 359: Is there a reason why we're using a 'custom' way to detect this instead of the helper function `doesObjectStorageExists`? If there is, we probably should get rid of the helper function and roll it into `verifyObjectStorageExists` and if there isn't, refactor it to be usable in both cases. 

It also seems a tad weird to me that we have 'doesExist' and 'doesNotExist' that do different things? Could this be just implemented as `doesNotExist() { return !doesExist() }` (similar to how bucketExists/bucketDoesNotExist are)? 

Or is it just a naming-does-not-follow-semantics kind of an issue - eg. should `doesObjectStorageExist` actually be something along the lines of `doesObjectStorageMinioInstanceExist`? (not familiar with the semantics, unfortunately)

upcloud/resource_upcloud_objectstorage_test.go
```diff
@@ -349,16 +349,20 @@ func verifyObjectStorageExists(accessKey, secretKey, name string) resource.TestC
...
+				UUID: rs.Primary.ID,
+			})
+
+			if err == nil {
```
moitias, line 363: Maybe we should also check that the error is what we actually expect (with [errors.Is](https://golang.org/pkg/errors/#Is) or just even err.Error() == 'expected string'), rather than just making sure it's not nil? 

Any other errors should also return some sort of an error to fail the test but it's a good practice to check the particular error that we expect this call to return if we depend on such behaviour.

moitias:
Some minor styilistic nitpicks but broadly LGTM :+1:

upcloud/resource_upcloud_objectstorage_test.go
kaminek: The reason of not using `doesObjectStorageExists` to check the s3 bucket availability is because of a race condition.
 the s3 endpoint is still availability few seconds after the API delete call, that's why we check against the API and not the resource. BTW terraform should always interract with the API not the resources.

kaminek:
There are 3 other failing tests for other resources, not related to this MR. I'll check them after

upcloud/resource_upcloud_objectstorage_test.go
moitias: Ahh, right - well, if we absolutely cant implement some sort of a waiting logic, then I would propose just naming these more clearly to indicate the difference, and maybe adding the reasoning in a comment to the source?

upcloud/resource_upcloud_objectstorage_test.go
LordChocoBlizzBallz: Requires state *terraform.State to work and this seems to be good way to do it.

kaminek:
Merging, the failing tests aren't related to this MR

kaminek (approved):

PR #142, opened 2021-05-10
chore(ci): add golangci linter

kaminek:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,14 @@
+
+name: golangci-lint
+on:
+  pull_request:
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.37.1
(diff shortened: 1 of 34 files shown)


moitias:
:raised_hands: LGTM :+1:

moitias (approved):

moitias (changes requested):
We should at least;

- enable go fmt in golangci-lint
- add pre-commit hooks that run the same linters

before merging

.github/workflows/lint.yml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v2
+        with:
+          version: v1.37.1
```
scop, line 14: If there's no reason to install an older one:

```suggestion
          version: v1.40.0
```

scop:
>     * enable go fmt in golangci-lint

Tested adding `gofmt` to the config locally, it resulted in no new findings. But agree we should add it if there's no strong/particular reason not to.

moitias (approved):

PR #144, opened 2021-05-17
fix(resource/router): fix creation of attachedNetworks for routers

moitias:
Previously, the initialization created a slice with len,cap=len(attachednetworks) filled with default values (eg. "" for strings).

After appending to the list, the list then had extra empty entries in the beginning ["", "uuid"] which were returned to TF. 

This is now fixed.


kaminek (approved):
nice catch, LGTM

kaminek:
The test error isn't related to this MR

PR #146, opened 2021-05-20
fix(server): prevent some attribute update from restarting

kaminek:
Some attributes don't need a restart to be updated

### DIFF
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -176,6 +176,7 @@ func resourceUpCloudServer() *schema.Resource {
 				Description: "The pricing plan used for the server",
 				Type:        schema.TypeString,
 				Optional:    true,
+				Computed:    true,
 			},
 			"storage_devices": {
 				Description: "A list of storage devices associated with the server",
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -225,6 +225,7 @@ func resourceUpCloudServer() *schema.Resource {
 						"size": {
 							Description:  "The size of the storage in gigabytes",
 							Type:         schema.TypeInt,
+							Computed:     true,
 							Optional:     true,
 							ValidateFunc: validation.IntBetween(10, 2048),
 						},
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -420,13 +420,12 @@ func resourceUpCloudServerUpdate(ctx context.Context, d *schema.ResourceData, me
 	if err != nil {
 		return diag.FromErr(err)
 	}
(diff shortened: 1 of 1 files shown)


upcloud/resource_upcloud_server.go
```diff
@@ -420,94 +429,111 @@ func resourceUpCloudServerUpdate(ctx context.Context, d *schema.ResourceData, me
...
 			}
 		}
+
+		// handle the other storage devices
+		if d.HasChange("storage_devices") {
+			o, n := d.GetChange("storage_devices")
+
+			// detach the devices that should be detached or should be re-attached with different parameters
+			for _, storageDevice := range o.(*schema.Set).Difference(n.(*schema.Set)).List() {
+				if serverDetails.StorageDevice(storageDevice.(map[string]interface{})["storage"].(string)) == nil {
```
moitias, line 504: Could we move this cast to a separate variable? It could make this a little bit readable with less `(map[string]interface{})`s all around.

moitias (approved):
Would like a minor readability improvement but otherwise LGTM :+1:

upcloud/resource_upcloud_server.go
```diff
@@ -420,94 +429,112 @@ func resourceUpCloudServerUpdate(ctx context.Context, d *schema.ResourceData, me
...
+				if _, err := client.DetachStorage(&request.DetachStorageRequest{
+					ServerUUID: d.Id(),
+					Address:    storageDevice.(map[string]interface{})["address"].(string),
+				}); err != nil {
+					return diag.FromErr(err)
+				}
+			}
+			// attach the storages that are new or have changed
+			for _, storageDevice := range n.(*schema.Set).Difference(o.(*schema.Set)).List() {
+				storageDevice := storageDevice.(map[string]interface{})
```
moitias, line 517: I like this approach better than the one above - eg. get the item from the list and cast it once and then pick the relevant attributes out of it.

However, using the same name here could be a bit dangerous due to the scope being maybe a bit confusing? Maybe something like;
```
	for _, rawStorageDevice := range n.(*schema.Set).Difference(o.(*schema.Set)).List() {
				storageDevice := rawStorageDevice.(map[string]interface{})
``` 
or, changedStorageDeviceData or something? Or; even better, make a helper that extracts (casted) added/removed lists from the changes? So something along the lines of;
```
func getStorageDeviceChanges(d *schema.ResourceData) (removedDevices []map[string]interface{}, addedDevices []map[string]interface{}) {
... 
}
``` 

But tbh, it's probably just me being grumpy in the afternoon - this works for me too if you'd rather jump on to something else, up to you!

kaminek: this seems to be more readable, I'll do it

kaminek: The issue is this way was implemented all over the codebase :disappointed:

PR #148, opened 2021-05-24
Improve Terraform 0.12 example

Darep:
Removes the version-attribute as it breaks things in 0.12, and adds link to the 0.12 build-section to make it clearer that it needs to be built separately for 0.12.


moitias:
LGTM :+1:

moitias (approved):

PR #149, opened 2021-05-24
Fix CHANGELOG

Darep:
- 2.1.0 is not released => changes to "Unreleased"
- adds links for versions
- adds notes about 1.0.0


moitias (approved):
LGTM :+1:

PR #150, opened 2021-05-24
feat(server): add tags attribute

kaminek:
Add tags attribute in server resource.
Since the API/SDK don't allow managing tags from /server resource, we need to use /tag resource. Basically on any creation/update we handle the tags separately then we add/remove them from server resource

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -220,3 +220,87 @@ func VerifyServerStarted(startRequest request.StartServerRequest, meta interface
 	}
 	return nil
 }
+
+// RemoveOldServerTags removes tags from server.
+// deletes the tag if no server left with the tag name
+func RemoveOldServerTags(service *service.Service, serverUUID string, tags []string) error {
+	if len(tags) == 0 {
+		return nil
+	}
+
+	if _, err := service.UntagServer(&request.UntagServerRequest{
+		UUID: serverUUID,
+		Tags: tags,
+	}); err != nil {
+		return err
+	}
+
+	currTags, err := service.GetTags()
+	if err != nil {
+		return err
+	}
+
+	// Remove unused tags
+	for _, currTag := range currTags.Tags {
+		for _, tag := range tags {
+			if tag == currTag.Name {
+				// remove tag from server if it's empty
+				if len(currTag.Servers) == 0 {
+					if err := service.DeleteTag(&request.DeleteTagRequest{
+						Name: currTag.Name,
+					}); err != nil {
+						return err
+					}
+				}
+			}
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


internal/server/server.go
```diff
@@ -220,3 +220,87 @@ func VerifyServerStarted(startRequest request.StartServerRequest, meta interface
 	}
 	return nil
 }
+
+// RemoveOldServerTags removes tags from server.
+// deletes the tag if no server left with the tag name
```
moitias, line 225: Are tags *only* related to servers (for now?) Seems a bit dangerous to me to auto-delete but not sure I understand the semantics/issues involved.

upcloud/resource_upcloud_server.go
```diff
@@ -299,24 +308,39 @@ func resourceUpCloudServerCreate(ctx context.Context, d *schema.ResourceData, me
 	if err != nil {
 		return diag.FromErr(err)
 	}
-	server, err := client.CreateServer(r)
+	s, err := client.CreateServer(r)
```
moitias, line 311: Is there a reason why we use `s` instead of the more descriptive `server` as before?

moitias (approved):
Minor nits, LGTM otherwise :+1:

internal/server/server.go
kaminek: yes the way tags are implemented for now are only related to server, that's why

upcloud/resource_upcloud_server.go
kaminek: server is used in here https://github.com/UpCloudLtd/terraform-provider-upcloud/pull/150/files#diff-932d31a4bcd943933a162fe0586b58bd2b829b2e5a3b9f81ab060c35c5cbcdceR322
which conflitcts

moitias: ahh, right, maybe just `serverdetails` or `createdserver`, then?

internal/server/server.go
moitias: :+1: Got it, makes sense, then. I'm  a bit concerned over what happens when/if they don't? Is there a downside to not deleting the tags? Do we have a way of querying the resources/servers of a tag - if we do and that seems like a generic way that should work in the future as well, could we use it to make sure? 

If it seems too complex/impossible to do now, maybe at the very least move the deletion to a separate method and make it explicit that this is the functionality now? Otherwise, figuring it out would need one to look into RemoveOldServerTags which doesn't really give an indication that it will also delete tags.

kaminek: removing the tags when there are no server on them is the best way of keeping them clean, I don't know that is the limit of tags resource in the API, but I think this is the best way to do it since there are also tag resources in tf that may conflict if the tags were deleted from server and wanted to be used as a standalone tag resource.

PR #151, opened 2021-05-25
fix(router): allow detaching router and deleting attached routers

### LINKED ISSUE #137: Cannot delete a router when it's attached
I had this definition:
```
resource "upcloud_network" "my-private-network" {
  name = "my-private-network"
  zone = "us-chi1"

  router = my-router.id

  ip_network {
    address            = "10.0.0.0/24"
    dhcp               = true
    family  = "IPv4"
  }
}

resource "upcloud_router" "my-router" {
  name = "my-router"
}
```
Then edited to this:
```
resource "upcloud_network" "my-private-network" {
  name = "my-private-network"
  zone = "us-chi1"

  ip_network {
    address            = "10.0.0.0/24"
    dhcp               = true
    family  = "IPv4"
  }
}

```

But it failed to cascade the changes in the right order:

```
Terraform will perform the following actions:

  # upcloud_network.my-private-network will be updated in-place
  ~ resource "upcloud_network" "my-private-network" {
        id      = "0366acb2-067a-48c3-b2d0-3fd4fb8b8846"
        name    = "my-private-network"
      - router  = "04bf929b-1ed0-4cc8-824d-c5215843ee34" -> null
        # (3 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

  # upcloud_router.my-router will be destroyed
  - resource "upcloud_router" "my-router" {
      - attached_networks = [] -> null
      - id                = "04bf929b-1ed0-4cc8-824d-c5215843ee34" -> null
      - name              = "my-router" -> null
    }

Plan: 0 to add, 1 to change, 1 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

upcloud_router.my-router: Destroying... [id=04bf929b-1ed0-4cc8-824d-c5215843ee34]
╷
│ Error: There are 1 networks attached to the router 04bf929b-1ed0-4cc8-824d-c5215843ee34. (ROUTER_ATTACHED)
│ 
│ 
╵
``` 
The expected behavior was to remove the router from the networks before trying to delete it.

moitias:
- pulls in go-api 4.0.0, allowing fot detaching to work
- implement the new go-api detachment methods
- make sure routers are detached when deleting

Fixes #137

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/UpCloudLtd/terraform-provider-upcloud
 go 1.14
 
 require (
-	github.com/UpCloudLtd/upcloud-go-api v0.0.0-20210127073406-2964ed7e5972
+	github.com/UpCloudLtd/upcloud-go-api v0.0.0-20210525102709-661419a1956b
 	github.com/agext/levenshtein v1.2.3 // indirect
 	github.com/fatih/color v1.10.0 // indirect
 	github.com/google/uuid v1.2.0 // indirect
--- a/upcloud/resource_upcloud_router_test.go
+++ b/upcloud/resource_upcloud_router_test.go
@@ -90,6 +90,109 @@ func TestAccUpCloudRouter_import(t *testing.T) {
 	})
 }
 
+func TestAccUpCloudRouter_detach(t *testing.T) {
+	var providers []*schema.Provider
+	var router upcloud.Router
+	var network upcloud.Network
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviderFactories(&providers),
+		CheckDestroy:      testAccCheckRouterNetworkDestroy,
+		Steps: []resource.TestStep{
+			{
+				// first create network and router attached
+				Config: testAccRouterNetworkConfig("testrouter", "testnetwork", true, true),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRouterExists("upcloud_router.terraform_test_router", &router),
+					testAccCheckNetworkExists("upcloud_network.terraform_test_network", &network),
+				),
+			},
+			{
+				ResourceName:      "upcloud_router.terraform_test_router",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateCheck: func(states []*terraform.InstanceState) error {
+					for _, s := range states {
+						if s.Attributes["attached_networks.#"] != "1" {
+							return fmt.Errorf("expected 1 network, got %v", s.Attributes["attached_networks.#"])
+						}
+					}
+					return nil
+				},
+			},
+			{
+				// and then change them to detached
+				Config: testAccRouterNetworkConfig("testrouter", "testnetwork", true, false),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRouterExists("upcloud_router.terraform_test_router", &router),
+					testAccCheckNetworkExists("upcloud_network.terraform_test_network", &network),
+				),
+			},
+			{
+				ResourceName:      "upcloud_router.terraform_test_router",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateCheck: func(states []*terraform.InstanceState) error {
+					for _, s := range states {
+						if s.Attributes["attached_networks.#"] != "0" {
+							return fmt.Errorf("expected 0 networks, got %v", s.Attributes["attached_networks.#"])
+						}
+					}
+					return nil
+				},
+			},
+		},
+	})
+}
+
+func TestAccUpCloudRouter_attachedDelete(t *testing.T) {
+	var providers []*schema.Provider
+	var router upcloud.Router
+	var network upcloud.Network
+	resource.Test(t, resource.TestCase{
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviderFactories(&providers),
+		CheckDestroy:      testAccCheckRouterNetworkDestroy,
+		Steps: []resource.TestStep{
+			{
+				// first create network and router attached
+				Config: testAccRouterNetworkConfig("testrouter", "testnetwork", true, true),
+				Check: resource.ComposeTestCheckFunc(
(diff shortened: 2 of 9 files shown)


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/UpCloudLtd/terraform-provider-upcloud
 go 1.14
 
 require (
-	github.com/UpCloudLtd/upcloud-go-api v0.0.0-20210127073406-2964ed7e5972
+	github.com/UpCloudLtd/upcloud-go-api v0.0.0-20210525102709-661419a1956b
```
kaminek, line 6: why this is not the actual tag?

upcloud/resource_upcloud_router_test.go
```diff
@@ -90,6 +90,109 @@ func TestAccUpCloudRouter_import(t *testing.T) {
...
+		PreCheck:          func() { testAccPreCheck(t) },
+		ProviderFactories: testAccProviderFactories(&providers),
+		CheckDestroy:      testAccCheckRouterNetworkDestroy,
+		Steps: []resource.TestStep{
+			{
+				// first create network and router attached
+				Config: testAccRouterNetworkConfig("testrouter", "testnetwork", true, true),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckRouterExists("upcloud_router.terraform_test_router", &router),
+					testAccCheckNetworkExists("upcloud_network.terraform_test_network", &network),
```
kaminek, line 107: can we check if the different computed attributes are set? with `resource.TestCheckResourceAttrSet()`.
Same every where were there are attributes set

upcloud/resource_upcloud_router_test.go
```diff
@@ -90,6 +90,109 @@ func TestAccUpCloudRouter_import(t *testing.T) {
...
+				ImportStateVerify: true,
+				ImportStateCheck: func(states []*terraform.InstanceState) error {
+					for _, s := range states {
+						if s.Attributes["attached_networks.#"] != "1" {
+							return fmt.Errorf("expected 1 network, got %v", s.Attributes["attached_networks.#"])
+						}
+					}
+					return nil
+				},
+			},
```
kaminek, line 122: This section can be replaced in the previous checks like
`resource.TestCheckResourceAttr("upcloud_router.testrouter", "name", "testrouter")`
same for number of networks and so one

upcloud/resource_upcloud_router_test.go
```diff
@@ -90,6 +90,109 @@ func TestAccUpCloudRouter_import(t *testing.T) {
...
+				ImportStateVerify: true,
+				ImportStateCheck: func(states []*terraform.InstanceState) error {
+					for _, s := range states {
+						if s.Attributes["attached_networks.#"] != "0" {
+							return fmt.Errorf("expected 0 networks, got %v", s.Attributes["attached_networks.#"])
+						}
+					}
+					return nil
+				},
+			},
```
kaminek, line 143: same here, this section can be tested above with `TestCheckResourceAttr()`

upcloud/resource_upcloud_router_test.go
```diff
@@ -90,6 +90,109 @@ func TestAccUpCloudRouter_import(t *testing.T) {
...
+				ImportStateVerify: true,
+				ImportStateCheck: func(states []*terraform.InstanceState) error {
+					for _, s := range states {
+						if s.Attributes["attached_networks.#"] != "1" {
+							return fmt.Errorf("expected 1 network, got %v", s.Attributes["attached_networks.#"])
+						}
+					}
+					return nil
+				},
+			},
```
kaminek, line 177: same as above

upcloud/resource_upcloud_router_test.go
```diff
@@ -153,9 +286,72 @@ func testAccCheckRouterDestroy(s *terraform.State) error {
...
+}
+
 func testAccRouterConfig(name string) string {
 	return fmt.Sprintf(`
 resource "upcloud_router" "my_example_router" {
   name = "%s"
 }`, name)
 }
+
+func testAccRouterNetworkConfig(routerName, networkName string, includeRouter, routerAttached bool) string {
```
kaminek, line 329: Not really fan of function that returns formated tf config, as the reader/reviewer cannot understand the test easly. up for debate

kaminek (changes requested):

go.mod
moitias: This is how it ended up with `go get github.com/UpCloudLtd/upcloud-go-api@4.0.0` - I'm guessing it's related to the `v0.0.0` prefix and my working theory based on that is that `go` (the tool) expects git tags to be `vX.Y.Z` rather than `X.Y.Z` but did not delve deeper into it for the time being.

upcloud/resource_upcloud_router_test.go
moitias: :+1: Excellent, thanks!

upcloud/resource_upcloud_router_test.go
moitias: Would you prefer full configs in the `Config`-field or what would be the preferred way of doing it? 

This way does have the advantage of cloning most of the configuration - eg. no need to change 3 configurations in the same way in three different steps but I do agree it's quite opaque and makes it quite tough to follow along, especially with the two bool parameters. Will rethink, at least, unless you have a great suggestion in mind?

go.mod
aakso: AFAIK the major version should be included as a suffix for the import path. So `github.com/UpCloudLtd/upcloud-go-api/v4`. Ref: https://golang.org/ref/mod

upcloud/resource_upcloud_router_test.go
moitias: This turned out to be difficult, due to a circular dependency/chicken-and-egg issue; 
- due to network being dependent on the router (due to the reference), router gets created first. at this point, the computed attribute (attached_networks) is empty.
- network gets created with attachment, but the router resource has no way of knowing of the change, so internally it still has an empty attached_networks
- this will not change until we reimport/update the router and there's no way to trigger such a reimport from another resource (due to the design of tf, which is probably due to wanting to avoid the possibility of circular dependencies)

kaminek (approved):

PR #152, opened 2021-05-26
fix(storage): check size before cloning a device

kaminek:


moitias (approved):
LGTM :+1:

PR #153, opened 2021-05-28
fix(storage): fix address formating

kaminek:
Use storage device address as bus type without the bus id


internal/utils/utils.go
```diff
@@ -82,3 +83,9 @@ func ExpandStrings(data interface{}) []string {
 
 	return strSlice
 }
+
+// StorageAddressFormat takes the address in any format and extracts the bus
+// type only (ide/scsi/virtio)
+func StorageAddressFormat(address string) string {
+	return strings.Split(address, ":")[0]
```
moitias, line 90: This can panic (at least with an empty string,  `strings.Split("",":")` returns `[]string{}`), so lets check and handle that case, at least.

moitias (changes requested):
Besides the panic, LGTM :+1:

PopoSensei (approved):
LGTM

PR #154, opened 2021-05-28
fix(server/login): make login user optional

kaminek:


moitias (approved):
Looks like the test failures are not related/transient, so LGTM :+1:

PR #155, opened 2021-05-28
chore(ci): update release workflow

kaminek:
[follow official doc](https://learn.hashicorp.com/tutorials/terraform/provider-release-publish?utm_offer=ARTICLE_PAGE)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,49 @@
+# This GitHub action can publish assets for release when a tag is created.
+# Currently its setup to run on any tag that matches the pattern "v*" (ie. v0.1.0).
+#
+# This uses an action (hashicorp/ghaction-import-gpg) that assumes you set your 
+# private key in the `GPG_PRIVATE_KEY` secret and passphrase in the `PASSPHRASE`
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+        with:
+          go-version: 1.16
+      -
+        name: Import GPG key
+        id: import_gpg
+        uses: hashicorp/ghaction-import-gpg@v2.1.0
+        env:
+          # These secrets will need to be configured for the repository:
+          GPG_PRIVATE_KEY: ${{ secrets.GPG_PRIVATE_KEY }}
+          PASSPHRASE: ${{ secrets.PASSPHRASE }}
```
Darep, line 39: Was the signing key & passphrase added to GitHub secrets already? Is they key a self-generated one?

kaminek: not yet I'll ask the ones that used to release manually

moitias (approved):
LGTM :+1:

.github/workflows/release.yml
Darep: Ah ok, so we'll use the same keys we used for releasing before :) I wonder where those are...

PR #165, opened 2021-06-09
fix(client): fix user-agent value

kaminek:


moitias (changes requested):
Should we also delete the old internal/globals.go, if we're not using it any more?

Besides that, LGTM :+1:

kaminek:
yes there is a tailing file/dir

moitias (approved):
LGTM, let's goooo!

PR #172, opened 2021-08-30
Fixes for local dev

Darep:
- fix the path being generated by `make build` (the version is supposed to be e.g. "2.1.1" but it was "v2.1.1" or "v2.1.1-dirty")
- add extra help to README

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -3,8 +3,9 @@ GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 
 MODULE   = $(shell env GO111MODULE=on go list -m)
-VERSION ?= $(shell git describe --tags --always --dirty --match=v* 2> /dev/null || \
+GIT_VERSION ?= $(shell git describe --tags --always --dirty --match=v* 2> /dev/null || \
 			cat $(CURDIR)/.version 2> /dev/null || echo v0)
+VERSION = $(shell echo $(GIT_VERSION) | sed 's/^v//' | sed 's/-.*//')
 
 PROVIDER_HOSTNAME=registry.upcloud.com
 PROVIDER_NAMESPACE=upcloud
@@ -18,7 +19,7 @@ build: fmtcheck
 	@mkdir -p $(PROVIDER_PATH)
 	go build \
 		-tags release \
-		-ldflags '-X $(MODULE)/internal/config.Version=$(VERSION)' \
+		-ldflags '-X $(MODULE)/internal/config.Version=$(GIT_VERSION)' \
 		-o $(PROVIDER_PATH)/terraform-provider-$(PROVIDER_NAMESPACE)_v$(VERSION)
 
 build_0_12: fmtcheck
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -5,12 +5,13 @@ WEBSITE_REPO=github.com/hashicorp/terraform-website
 MODULE   = $(shell env GO111MODULE=on go list -m)
 VERSION ?= $(shell git describe --tags --always --dirty --match=v* 2> /dev/null || \
 			cat $(CURDIR)/.version 2> /dev/null || echo v0)
+VERSION_NUMBER = $(shell echo $(VERSION) | sed 's/v//' | sed 's/-dirty//')
```
ka-myl, line 8: Probably don't need to do `| sed 's/-dirty//'` if you removed the `--dirty` flag in the line above 🤔

GNUmakefile
```diff
@@ -19,7 +20,7 @@ build: fmtcheck
 	go build \
 		-tags release \
 		-ldflags '-X $(MODULE)/internal/config.Version=$(VERSION)' \
```
ka-myl, line 22: Need to update to `VERSION_NUMBER` here too

Darep: This is for the user-agent and imo it should have the v and the -dirty 😄

GNUmakefile
Darep: The dirty thing feels ok to me

README.md
```diff
@@ -251,6 +251,21 @@ value specified in the makefile and in this case the version is 2.0.0.
...
 After the provider has been built you can then use standard terraform commands.
 
+Use this provider config with the local version:
+
+```
+terraform {
+  required_providers {
+    upcloud = {
+      source = "registry.upcloud.com/upcloud/upcloud"
+      version = "~> 2.0"
```
kaminek, line 261: please promote the latest version 2.1

GNUmakefile
kaminek: I think that the original `VERSION` should be `GIT_VERSION` and this new one should be `VERSION`

kaminek (changes requested):

GNUmakefile
Darep: Will change that! 👍

README.md
Darep: Will do!

GNUmakefile
```diff
@@ -3,8 +3,9 @@ GOFMT_FILES?=$$(find . -name '*.go' |grep -v vendor)
 WEBSITE_REPO=github.com/hashicorp/terraform-website
 
 MODULE   = $(shell env GO111MODULE=on go list -m)
-VERSION ?= $(shell git describe --tags --always --dirty --match=v* 2> /dev/null || \
+GIT_VERSION ?= $(shell git describe --tags --always --dirty --match=v* 2> /dev/null || \
 			cat $(CURDIR)/.version 2> /dev/null || echo v0)
+VERSION = $(shell echo $(GIT_VERSION) | sed 's/^v//' | sed 's/-.*//')
```
Darep, line 8: note: changed last sed from -dirty to -.* as we can also get git versions like `v2.1.1-2-edbe3a`

kaminek (approved):

PR #173, opened 2021-09-03
Change server/tags from List to Set

mikegager:
I was encountering an annoyance where some of my server/tags would show changed with each plan and be removed and readded with each apply. It would complete the job successfully, just that it should not have been reporting changes being required each run.

I found that server/tags is defined as a TypeList, where it seemed it should be a TypeSet, which appears to be what the deprecated tags resource used.

This patch changes server/tags to be defined as a TypeSet, and updates the ExpandStrings utility method to work with the Set. This change resolved my issue and I did not notice any side-effects in my testing (server create, add tag, remove tag, server delete).

The change seems pretty trivial to me, but this is my first time editing Go, so I can only say that it appears to work, not that it is correct.

Thank you.

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -77,7 +77,7 @@ func WithRetry(fn func() (interface{}, error), retries int, delay time.Duration)
 // ExpandStrings expands a terraform interface to slice of str
 func ExpandStrings(data interface{}) []string {
 	strSlice := []string{}
-	for _, s := range data.([]interface{}) {
+	for _, s := range data.(*schema.Set).List() {
 		strSlice = append(strSlice, s.(string))
 	}
 
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -81,7 +81,7 @@ func resourceUpCloudServer() *schema.Resource {
 			},
 			"tags": {
 				Description: "The server related tags",
-				Type:        schema.TypeList,
+				Type:        schema.TypeSet,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
 				},


kaminek:
Hi @mikegager
Thanks for the effort, but I cannot reproduce the behavior you described above (even in the tests), can you please share your config?

mikegager:
Thank you for having a look at this, and I'm sorry for the delay getting back to you (I've been swamped lately). I was afraid that this might be something relating to something specific to my account or its history (I only recently started here, trying to move into terraform where previously it had generally been managed through the WebUI). The odd part being that it appears to only happen with certain tags, though I haven't been able to find a meaningful pattern to them (and at one point tried removing one of them from everything, seeing that it didn't try to re-tag anything, then re-added it).

I am able to reproduce it on my setup with the attached minimal terraform. 
[main.txt](https://github.com/UpCloudLtd/terraform-provider-upcloud/files/7192956/main.txt)
Note that in this example, it appears to only happen for the "stg" tag, and if I change the "env" variable to some random string instead, it appears to work as expected.

I'll also mention, as it is likely somewhat related, my normal setup is now a bit more elaborate using some modules to organize things. I will generally get some TAG_NOT_FOUND error messages for some of the tags, but AFAICT they don't cause any trouble finishing out the run, and don't obviously align with the ones that it wants to recreate on the following run.

Being that most of the changes look like the one pasted below, my first thought was that it had something to do with ordering, which is what got me thinking of making it a set rather than a list. Adding a toset() to the concat() in the terraform file was insufficient, so went looking a bit deeper for where I could try forcing it to be a set.  I did some basic testing and it appeared to work, though I still haven't spent enough time with the "fixed" version to be super-confident in it.

      ~ tags     = [
          - "stg",
            "neverused11",
            "neverused12",
          + "stg",
        ]

Given the circumstances, I certainly will not be offended if you would like to reject this as unreproducible, especially as it seems to just be me, and I haven't been able to pin down the patterns involved yet. Thanks again.

kangasta:
Here is a simple config that I was able to reproduce this with:

```terraform
terraform {
  required_providers {
    upcloud = {
      source  = "registry.upcloud.com/upcloud/upcloud"
      version = "~> 2.4"
    }
  }
}

resource "upcloud_server" "tag-test-34" {
  zone     = "pl-waw1"
  hostname = "tag-test-34"
  tags = [
    "tag-3",
    "tag-4"
  ]

  template {
    storage = "01000000-0000-4000-8000-000030200200"
    size    = 10
  }

  network_interface {
    type = "public"
  }
}

resource "upcloud_server" "tag-test-1234" {
  zone     = "pl-waw1"
  hostname = "tag-test-1234"
  tags = [
    # Run terraform apply once, uncomment these after that, and run terraform apply twice again
    # "tag-1",
    # "tag-2",
    "tag-3",
    "tag-4"
  ]

  template {
    storage = "01000000-0000-4000-8000-000030200200"
    size    = 10
  }

  network_interface {
    type = "public"
  }
}

```

kangasta:
Thanks again for reporting this and proposing an solution. This might be a breaking change, so we will probably have to wait until next major version before merging this in. Before that, changes from #257 will supress diff if only the order of the tags changes.

kangasta:
I'll close this as completed: the proposed changes were cherry-picked into #414 and have been since released

PR #175, opened 2021-09-03
Feat/simple backups support

ka-myl:
Adds support for simple backup option for servers.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -320,3 +320,36 @@ func UpdateServerTags(service *service.Service, serverUUID string, oldTags, newT
 
 	return AddNewServerTags(service, serverUUID, newTags)
 }
+
+// ValidateServerAndStorageBackupRules checks if the server and (to be) attached storages have conflicting backup rules.
+func ValidateServerAndStorageBackupRules(service *service.Service, d *schema.ResourceData) error {
+	if _, ok := d.GetOk("simple_backup"); ok {
+		if attr, ok := d.GetOk("storage_devices"); ok {
+			sds := attr.(*schema.Set).List()
+
+			for _, i := range sds {
+				e := i.(map[string]interface{})
+				id := e["storage"]
+
+				sd, err := service.GetStorageDetails(&request.GetStorageDetailsRequest{
+					UUID: id.(string),
+				})
+
+				if err != nil {
+					return err
+				}
+
+				if sd.BackupRule != nil && sd.BackupRule.Time != "" {
+					return fmt.Errorf("invalid configuration; server with simple backup schedule cannot have storages with their own backup rule attached (check storage %s - %s)", sd.Title, sd.UUID)
+				}
+			}
+		}
+
+		// If there is no storage_devices block - no need to do anything. Also no need to check the template
+		// as setting simple_backup option for the server and backup_rule for the template is not allowed on the schema level
+		return nil
+	}
+
+	// If no simple backup option is set for server - no need to do anything
+	return nil
+}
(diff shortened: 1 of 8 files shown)


internal/server/server.go
```diff
@@ -304,3 +320,36 @@ func UpdateServerTags(service *service.Service, serverUUID string, oldTags, newT
...
+func ValidateServerAndStorageBackupRules(service *service.Service, d *schema.ResourceData) error {
+	if _, ok := d.GetOk("simple_backup"); ok {
+		if attr, ok := d.GetOk("storage_devices"); ok {
+			sds := attr.(*schema.Set).List()
+
+			for _, i := range sds {
+				e := i.(map[string]interface{})
+				id := e["storage"]
+
+				sd, err := service.GetStorageDetails(&request.GetStorageDetailsRequest{
```
ka-myl, line 334: Doing a request for each storage is imo a bit heavy, but I haven't found a better way to validate this. If you have some other ideas please let me know

internal/server/server.go
```diff
@@ -304,3 +320,36 @@ func UpdateServerTags(service *service.Service, serverUUID string, oldTags, newT
 
 	return AddNewServerTags(service, serverUUID, newTags)
 }
+
+// ValidateServerAndStorageBackupRules checks if the server and (to be) attached storages have conflicting backup rules.
+func ValidateServerAndStorageBackupRules(service *service.Service, d *schema.ResourceData) error {
+	if _, ok := d.GetOk("simple_backup"); ok {
```
Darep, line 326: personally would reverse the check here so we immediately return nil in this case (so we don't have so many blocks/so much indentation)

ka-myl: Removed this whole function. It could lead to errors on subsequent updates with some configs.

kaminek:
@ka-myl can you rebase please and fix the conflicts?

upcloud/resource_upcloud_server.go
```diff
@@ -305,6 +307,29 @@ func resourceUpCloudServer() *schema.Resource {
 					},
 				},
 			},
+			"simple_backup": {
+				Description:   "Simple backup schedule configuration",
+				Type:          schema.TypeList,
```
kaminek, line 312: please use `TypeSet` that represents unordered collection

upcloud/resource_upcloud_server.go
```diff
@@ -385,6 +411,17 @@ func resourceUpCloudServerRead(ctx context.Context, d *schema.ResourceData, meta
...
 		_ = d.Set("firewall", false)
 	}
+	if server.SimpleBackup != "no" {
+		p := strings.Split(server.SimpleBackup, ",")
+		simpleBackup := map[string]interface{}{
+			"time": p[0],
+			"plan": p[1],
+		}
+
+		_ = d.Set("simple_backup", []map[string]interface{}{simpleBackup})
```
kaminek, line 421: just `simpleBackup`

upcloud/resource_upcloud_server.go
```diff
@@ -385,6 +411,17 @@ func resourceUpCloudServerRead(ctx context.Context, d *schema.ResourceData, meta
...
+	} else {
```
kaminek, line 422: not needed, default is `nil`

upcloud/resource_upcloud_server.go
```diff
@@ -390,6 +416,15 @@ func resourceUpCloudServerRead(ctx context.Context, d *schema.ResourceData, meta
 	} else {
 		_ = d.Set("firewall", false)
 	}
+	if server.SimpleBackup != "no" {
```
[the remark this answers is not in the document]
moitias: Yep, that's more of a -libs thing I feel!

PR #176, opened 2021-09-20
enhance test duration with parallelism

kaminek:
* Use go test framework parallelism to enhance test duration from more than 2h to 10m.
* Some tests are left because of compatibility issues, TBD
* Adapt resources params to accept parallelism


kaminek:
It seems that the Github runners are very slow, running the acceptance tests locally took about 11mins where here it's 19.
> with the same number of threads/workers (8)

upcloud/resource_upcloud_server_test.go
```diff
@@ -456,6 +457,7 @@ func testAccServerNetworkInterfaceConfig(nis ...networkInterface) string {
 	`)
 
 	for i, ni := range nis {
+		netID := rand.Intn(255)
```
moitias, line 460: This seems like a recipe for flaky tets :|

kaminek: this is only called by one test so, it's random only for one kind of test.

PR #178, opened 2021-09-22
fix(server): prevent empty tags from replanning

kaminek:
Given a server without tags, the SDK returns an empty array tag for that server. Terraform local state is that case is a nil tag array. So trying a replan will show a diff between the returned SDK values and the local state.

That patch considers an empty tag array as nil.


moitias:
Maybe add a test for the 'other way around' option as well, eg. what if the input has `tags:[]`?  Otherwise, LGTM :+1:

moitias (approved):

kaminek:
> Maybe add a test for the 'other way around' option as well, eg. what if the input has `tags:[]`? Otherwise, LGTM +1

Terraform considers tags: [] as nil, so nothing changes actually, if we decide to add a test for that we are actually testing terraform behavior and not the provider.

PR #180, opened 2021-09-23
fix(tests): disable parallelism for router tests

kaminek:
Remove parallelism for router related test because the API not stable for that endpoint and concurrent creation may end up with a 500.


moitias (approved):
Please add comments noting why these are not parallelized, otherwise LGTM :+1:

Issue #28, opened 2018-06-21
Build fails


jakolehm:
```
$ make build
==> Checking that code complies with gofmt requirements...
dep ensure
go install
can't load package: package github.com/UpCloudLtd/terraform-provider-upcloud: code in directory /Users/xxx/go/src/github.com/UpCloudLtd/terraform-provider-upcloud expects import "github.com/vtorhonen/terraform-provider-upcloud"
make: *** [build] Error 1
```

iler:
@jakolehm this is now fixed in #29

PR #187, opened 2021-10-22
feat(resource): implement managed database resource types

aakso:
This patch adds initial support for our Managed Databases. Moreover, we switch to importing the major >1 version of UpCloud GO SDK properly.


docs/resources/managed_database_mysql.md
```diff
@@ -0,0 +1,155 @@
...
+- **username** (String) User name for authentication with the server where to migrate data from
+
+
+
+<a id="nestedatt--components"></a>
+### Nested Schema for `components`
+
+Read-Only:
+
+- **component** (String)
```
ka-myl, line 139: Should we include some descriptions here? Or is it something that a mysql-versed person should know anyway?

peknur: Good catch thanks! It seems that doc generator ignores `components` and also `node_states` field descriptions. I'll check if there is something wrong with schema or is this just doc gen. bug.

peknur: This seems to be known bug in tfplugindocs [#28](https://github.com/hashicorp/terraform-plugin-docs/issues/28). Those are read-only informational fields. I think that  in the future we can handle documentation in separate PR if needed.

docs/resources/managed_database_logical_database.md
```diff
@@ -0,0 +1,43 @@
...
+```terraform
+resource "upcloud_managed_database_postgresql" "example" {
+  name  = "postgres"
+  plan  = "1x1xCPU-2GB-25GB"
+  title = "postgres"
+  zone  = "fi-hel1"
+}
+
+resource "upcloud_managed_database_logical_database" "example_db" {
+  service = upcloud_managed_database_postgresql.example.id
```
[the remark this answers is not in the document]
peknur: This example creates (logical) database to PostgreSQL service (resource defined above) you can add extra databases to MySQL also. Anyway good point I'll add MySQL example also so there is no confusion.

ka-myl (approved):
LGTM

PR #192, opened 2021-11-09
Feat/server title support

peknur:
Add support for server title field

### DIFF
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -20,7 +20,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
-var cloudServerTitleLength int = 64
+const cloudServerTitleLength int = 64
 
 func resourceUpCloudServer() *schema.Resource {
 	return &schema.Resource{
@@ -761,7 +761,7 @@ func resourceUpCloudServerDelete(ctx context.Context, d *schema.ResourceData, me
 }
 
 func cloudServerDefaultTitleFromHostname(hostname string) string {
-	suffix := " (managed by terraform)"
+	const suffix string = " (managed by terraform)"
 	if len(hostname)+len(suffix) > cloudServerTitleLength {
 		hostname = fmt.Sprintf("%s...", hostname[:cloudServerTitleLength-len(suffix)-3])
 	}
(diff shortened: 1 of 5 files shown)


upcloud/resource_upcloud_server.go
```diff
@@ -400,7 +409,11 @@ func resourceUpCloudServerRead(ctx context.Context, d *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 	_ = d.Set("hostname", server.Hostname)
-	_ = d.Set("title", server.Title)
+	if server.Title != cloudServerDefaultTitleFromHostname(server.Hostname) {
+		_ = d.Set("title", server.Title)
+	} else {
+		_ = d.Set("title", nil)
```
Darep, line 415: What's the effect of setting this to nil here? Will TF make sure it's the default one?

upcloud/resource_upcloud_server.go
```diff
@@ -740,3 +759,11 @@ func resourceUpCloudServerDelete(ctx context.Context, d *schema.ResourceData, me
 
 	return diags
 }
+
+func cloudServerDefaultTitleFromHostname(hostname string) string {
+	suffix := " (managed by terraform)"
+	if len(hostname)+len(suffix) > cloudServerTitleLength {
+		hostname = fmt.Sprintf("%s...", hostname[:cloudServerTitleLength-len(suffix)-3])
```
Darep, line 766: UpCloud API accepts the single character ellipsis `…`
Could save two characters by using that here 😉

upcloud/resource_upcloud_server.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
+var cloudServerTitleLength int = 64
```
Darep, line 23: should this be `const`?

upcloud/resource_upcloud_server.go
```diff
@@ -740,3 +759,11 @@ func resourceUpCloudServerDelete(ctx context.Context, d *schema.ResourceData, me
 
 	return diags
 }
+
+func cloudServerDefaultTitleFromHostname(hostname string) string {
+	suffix := " (managed by terraform)"
```
Darep, line 764: Could this be a `const` above as well(?). Or is there a problem with placing the same const object in a bunch of strings 🤔

Darep (approved):
Did not test locally, but the changes look good to me! Note that I'm not a Go/Terraform expert 😉 

Dropped some questions/notes, check those maybe :blush:

upcloud/resource_upcloud_server.go
ka-myl: This sets the `title` value in `tfstate` to empty string. And I'm not sure if this is actually the desired behaviour. `title` field is marked as computed, so if I don't provide one, I would expect the computed value to end up in my tfstate.

ka-myl (commented):
I don't think we need a separate acceptance test for this feature, but perhaps it would be good to add title (and checks for it) to the `TestUpcloudServer_basic`?

PR #193, opened 2021-11-12
fix(server): fix custom plan updates (cpu/mem)

peknur:


upcloud/resource_upcloud_server_test.go
```diff
@@ -13,6 +13,59 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
 
+func TestUpcloudServer_custom(t *testing.T) {
```
ka-myl, line 16: Would name it `customPlan` or something like that, as users can also have custom server images :)

ka-myl (approved):
LGTM :+1:

upcloud/resource_upcloud_server_test.go
peknur: Good point! I will rename it.

Darep (approved):
LGTM! Tested locally and everything looked good! :) Fix works and other things still work, too :D

PR #195, opened 2021-11-17
feat(server): validate server hostname

peknur:
Validate hostname field to catch errors in plan phase.

### DIFF
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -866,7 +866,7 @@ func serverValidateHostname(hostname string) error {
 	}
 
 	if labelLen > labelMaxLen {
-		return fmt.Errorf("%s label is not in the range 1 - 63", hostname)
+		return fmt.Errorf("%s label is not in the range %d - %d", hostname, minLen, labelMaxLen)
 	}
 
 	if !nonNumeric {
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -33,10 +33,10 @@ func resourceUpCloudServer() *schema.Resource {
 		},
 		Schema: map[string]*schema.Schema{
 			"hostname": {
-				Description:  "A valid domain name",
-				Type:         schema.TypeString,
-				Required:     true,
-				ValidateFunc: validation.StringLenBetween(1, 128),
+				Description:      "A valid domain name",
+				Type:             schema.TypeString,
+				Required:         true,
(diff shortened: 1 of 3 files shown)


upcloud/resource_upcloud_server.go
```diff
@@ -767,3 +767,111 @@ func serverDefaultTitleFromHostname(hostname string) string {
...
+				return fmt.Errorf("'%s' character before hyphen cannot be dot", hostname[0:i+1])
+			}
+			labelLen++
+			nonNumeric = true
+		case c == '.':
+			if last == '.' || last == '-' {
+				return fmt.Errorf("'%s' character before dot cannot be dot or hyphen", hostname[0:i+1])
+			}
+			if labelLen > labelMaxLen || labelLen == 0 {
+				return fmt.Errorf("'%s' label is not in the range %d - %d", hostname[0:i+1], minLen, labelMaxLen)
```
ka-myl, line 859: If I'm reading this correctly - this will print the entire hostname, regardless of which label the loop is on. No biggie, but you might want to tweak the wording here a little.

upcloud/resource_upcloud_server.go
```diff
@@ -767,3 +767,111 @@ func serverDefaultTitleFromHostname(hostname string) string {
...
+			}
+			labelLen = 0
+		default:
+			return fmt.Errorf("%s contains illegal characters", hostname)
+		}
+		last = c
+	}
+
+	if labelLen > labelMaxLen {
+		return fmt.Errorf("%s label is not in the range 1 - 63", hostname)
```
ka-myl, line 869: Since we already have `labelMaxLen` constant, maybe it should be also used here, instead of hardcoding `63`? Again, not a blocker

ka-myl (approved):
LGTM 🚢

upcloud/resource_upcloud_server.go
peknur: Thank you @ka-myl for review!

At this point `hostname[0:i+1]` can also be partial hostname so it might not be label nor hostname :) 

Validation error can something like:
```
Error: Hostname validation failed
 
   with upcloud_server.example,
   on upcloud.tf line XX, in resource "upcloud_server" "example":
   21:   hostname = "example.exampleexampleexampleexampleexampleexampleexampleexampleexampleexample.com"
 
 'example.exampleexampleexampleexampleexampleexampleexampleexampleexampleexample.' label is not in the range 1 - 63
```

Darep (approved):
LGTM! Tested locally and all good 👍

PR #197, opened 2021-11-29
feat(server): validate plan and zone

peknur:
Validate field values before executing API commands.

### DIFF
--- a/upcloud/resource_upcloud_server.go
+++ b/upcloud/resource_upcloud_server.go
@@ -341,6 +341,14 @@ func resourceUpCloudServer() *schema.Resource {
 func resourceUpCloudServerCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	client := meta.(*service.Service)
 
+	if err := serverValidatePlan(client, d.Get("plan").(string)); err != nil {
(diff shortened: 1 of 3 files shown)


upcloud/resource_upcloud_server.go
```diff
@@ -542,7 +557,7 @@ func resourceUpCloudServerUpdate(ctx context.Context, d *schema.ResourceData, me
 	}
 
 	// Stop the server if the requested changes require it
-	if d.HasChanges("plan", "cpu", "mem", "template.0.size", "storage_devices") {
+	if d.HasChanges("cpu", "mem", "template.0.size", "storage_devices") || planHasChange {
```
ka-myl, line 560: Not sure if this is really needed. To me it's more readable when I have all the fields in one list, when I see the `||` I kind of assume this is a completely new condition. But that's just a style decision, not a blocker.

ka-myl (approved):
LGTM

upcloud/resource_upcloud_server.go
peknur: Yep I agreee that using `HasChanges` is probably more readable but I decided to use separate variable because `HasChanges` seems to be relatively big operations so caching it made sense.

PR #200, opened 2021-12-11
Fix/remove network servers field

ka-myl:
This PR removes the `servers` schema field for network resource. The field does not seem to be used anyway, attaching server to a network is done via servers network interfaces.


peknur (approved):
Looks good to me. Lets implement this when/if needed.

Darep (approved):

PR #201, opened 2021-12-13
fix(firewall): change firewall port types

peknur:
Change port types from int to string to avoid having zero values in state when importing rules with undefined port number(s).

Integer fields defined inside list element gets default value `0` in state file if not defined. This will cause planner to always suggest  update from `0` to `null` when field is not defined in config. Representing "undefined port" as `0` is not good practice since port `0` can have some use cases and it's also reserved by IANA.


upcloud/resource_upcloud_firewall_rules_test.go
```diff
@@ -312,3 +313,26 @@ func testUpcloudFirewallRulesInstanceConfigUpdate() string {
...
+
+	if diag := firewallRuleValidateOptionalPort("65535", p); len(diag) > 0 {
+		t.Error(diag[0].Detail)
+	}
+
+	if diag := firewallRuleValidateOptionalPort("abc", p); len(diag) < 1 {
+		t.Error("firewallRuleValidateOptionalPort failed 'abc' is not valid port")
+	}
+
+	if diag := firewallRuleValidateOptionalPort("0", p); len(diag) > 1 {
```
ka-myl, line 331: shouldn't this be `len(diag) < 1` ?

peknur: Yes it should :) I'll fix that.

ka-myl (approved):
LGTM 👍

PR #203, opened 2021-12-14
fix(firewall): Remove proto field's default value "tcp"

peknur:
Remove proto field's default value "tcp" as this prevents settings optional fields value to null and update validator to accept empty string which corresponds to any protocol.


ka-myl (approved):
👍

PR #204, opened 2021-12-21
fix(docs): Fix auto-generated docs

peknur:
Use tfplugindocs to generate docs using examples and resource descriptions.
Added also a new Make target "docs" (make docs).


docs/resources/storage.md
```diff
@@ -150,13 +129,6 @@ Required:
 - **retention** (Number) The number of days before a backup is automatically deleted
 - **time** (String) The time of day when the backup is created
 
-Notes:
```
ka-myl, line 153: Is there some way of adding such notes to the output? Via comment or something? If not, then we should move this information to examples somehow I guess

ka-myl:
Is there some way we can remove computed-only properties from the docs output? For example server resource docs specify that `id` is an optional property, when in fact user cannot set it. Would be cool to filter those out

docs/resources/storage.md
peknur: Missed those notes :) I created new template for server and added those notes to template.

peknur:
> Is there some way we can remove computed-only properties from the docs output? For example server resource docs specify that `id` is an optional property, when in fact user cannot set it. Would be cool to filter those out

I think that we cannot do that at the moment with tfplugindocs.

ka-myl (approved):
LGTM

Issue #181, opened 2021-09-30
Order of object storage buckets and their names incorrectly trigger changes


codeandroid:
It seems that the sort order of bucket names somehow influence the detection of changes.
If the order of buckets in the terraform definition is not ordered by names (ascending), then it will seemingly trigger changes (which get never applied because renames to buckets are not applied, anyway).

Create the object storage with this definition:

    resource "upcloud_object_storage" "test-issue" {
        size = 250
        name = "test-issue"
        zone = "de-fra1"
        access_key = var.UPCLOUD_STORAGE_ACCESS_KEY
        secret_key = var.UPCLOUD_STORAGE_SECRET_KEY
        description = "test"
        bucket {
            name = "ccc"
        }
        bucket {
            name = "aaa"
        }
        bucket {
            name = "bbb"
        }
    }

Then `terraform plan` results in:

    Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
    ~ update in-place

    Terraform will perform the following actions:

    # upcloud_object_storage.test-issue will be updated in-place
    ~ resource "upcloud_object_storage" "test-issue" {
            id          = "0000000-1111-2222-3333-444444444444"
            name        = "test-issue"
            # (9 unchanged attributes hidden)

        ~ bucket {
            ~ name = "aaa" -> "ccc"
            }
        ~ bucket {
            ~ name = "bbb" -> "aaa"
            }
        ~ bucket {
            ~ name = "ccc" -> "bbb"
            }
        }

    Plan: 0 to add, 1 to change, 0 to destroy.

Running `terraform apply` will *not apply* these changes. The buckets are not renamed.

Reordering the buckets in sort order like this:

    bucket {
        name = "aaa"
    }
    bucket {
        name = "bbb"
    }
    bucket {
        name = "ccc"
    }

... will result in the expected output for `terraform plan`:

    No changes. Your infrastructure matches the configuration.

    Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.

vtorhonen:
Fix is in and will be part of the next upcoming release. Thanks for reporting this @codeandroid!

PR #209, opened 2022-01-05
Use set instead of list for storing buckets in state

vtorhonen:
Currently buckets are created in the order of configuration. However they are stored in the order what the UpCloud API returns. Since users should be able to change the order of the configuration, and the resources are not dependent on each other, we should store the buckets into sets instead of lists.

Fixes issue #181.


vtorhonen:
Fix for tests at #211. Will rebase this branch after fix is merged.

peknur (approved):
Looks good to me :)

PR #211, opened 2022-01-07
Use unique storage object names in acceptance tests

vtorhonen:
Fixes an issue where acceptance tests were failing due `OBJECT_STORAGE_EXISTS` due to non-unique storage object name. Acceptance tests will autogenerate the storage object names with a common prefix. This prefix is also read by the [sweeper](https://www.terraform.io/plugin/sdkv2/testing/acceptance-tests/sweepers).

While modifying the file also changed some of the test related variables to be more descriptive.

### DIFF
--- a/upcloud/resource_upcloud_objectstorage_test.go
+++ b/upcloud/resource_upcloud_objectstorage_test.go
@@ -33,7 +33,6 @@ func init() {
 	resource.AddTestSweepers("object_storage_cleanup", &resource.Sweeper{
 		Name: "object_storage_cleanup",
 		F: func(region string) error {
-
 			username, ok := os.LookupEnv("UPCLOUD_USERNAME")
 			if !ok {
 				return fmt.Errorf("UPCLOUD_USERNAME must be set for acceptance tests")
--- a/upcloud/resource_upcloud_objectstorage_test.go
+++ b/upcloud/resource_upcloud_objectstorage_test.go
@@ -15,26 +15,24 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/minio/minio-go/v7"
+	"github.com/rs/xid"
 )
 
 const expectedDescription = "My object storage"
 const expectedZone = "pl-waw1"
 const expectedKey = "an access key"
 const expectedSecret = "a secret key"
+const runPrefix = "testacc-"
 
-const expectedName1 = "test-name1"
-const expectedName2 = "test-name2"
-const expectedName3 = "test-name3"
+var runID = xid.New().String()
+var expectedName1 = fmt.Sprintf("%s%s-1", runPrefix, runID)
+var expectedName2 = fmt.Sprintf("%s%s-2", runPrefix, runID)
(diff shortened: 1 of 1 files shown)


upcloud/resource_upcloud_objectstorage_test.go
```diff
@@ -15,27 +15,24 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/minio/minio-go/v7"
+	"github.com/rs/xid"
```
peknur, line 18: Just wondering do we need another dependency for this?  Could we just use `time.Now().Unix()` or `time.Now().UnixNano()`  to suffix variables?

upcloud/resource_upcloud_objectstorage_test.go
```diff
@@ -15,27 +15,24 @@ import (
...
 
-const expectedName1 = "test-name1"
-const expectedName2 = "test-name2"
-const expectedName3 = "test-name3"
+var runID = xid.New().String()
+var expectedName1 = fmt.Sprintf("%s%s-1", runPrefix, runID)
+var expectedName2 = fmt.Sprintf("%s%s-2", runPrefix, runID)
+var expectedName3 = fmt.Sprintf("%s%s-3", runPrefix, runID)
 
 func init() {
```
peknur, line 32: This is not part of changes but isn't this `init()` run every time our provider runs because our tests are in `upcloud` package ?  If so maybe we could rename it to eg `initObjectStorageTests()` and call it on each test function ?

upcloud/resource_upcloud_objectstorage_test.go
```diff
@@ -15,27 +15,24 @@ import (
...
 const expectedZone = "pl-waw1"
 const expectedKey = "an access key"
 const expectedSecret = "a secret key"
+const runPrefix = "testacc-"
 
-const expectedName1 = "test-name1"
-const expectedName2 = "test-name2"
-const expectedName3 = "test-name3"
+var runID = xid.New().String()
+var expectedName1 = fmt.Sprintf("%s%s-1", runPrefix, runID)
```
peknur, line 28: Should we prefix these `upcloud` package level variables with eg `objectStorageTest..` ? Not a big deal but might be confusing when editor suggests `expectedName1` when working on eg server resource :)

upcloud/resource_upcloud_objectstorage_test.go
vtorhonen: It's actually already a dependency (although an indirect one) at `go.mod`. But in case of leftovers from previous runs it would likely make things easier if you'd get some idea when the storage was created (already from the name). I'll change this to a timestamp. 👍

upcloud/resource_upcloud_objectstorage_test.go
vtorhonen: Good question. It'll run only when the `go test` is being run, and after all other `init()` calls have finished on non-test files. The Go compiler will ignore any `*_test.go` files from all packages so this piece of code won't affect the runtime.

upcloud/resource_upcloud_objectstorage_test.go
vtorhonen: Yes 👍

PR #213, opened 2022-01-13
fix(dbaas): increase dns retry interval

peknur:
Increase dns retry interval to match db.upclouddatabases.com SOA record.


vtorhonen (approved):
LGTM!

PR #215, opened 2022-01-15
refactor(dbaas): simplify validation of managed database related schemas

ka-myl:
This PR just replaces custom validation functions for dbaas related resources with functions from `validation` package


upcloud/resource_upcloud_managed_database_logical_database.go
```diff
@@ -54,20 +54,20 @@ func schemaUpCloudManagedDatabaseLogicalDatabase() map[string]*schema.Schema {
 			ForceNew:    true,
 		},
 		"character_set": {
-			Description:      "Default character set for the database (LC_CTYPE)",
-			Type:             schema.TypeString,
-			Optional:         true,
-			Computed:         true,
-			ForceNew:         true,
-			ValidateDiagFunc: validateManagedDatabaseLocale,
```
peknur, line 62: I think that we should use `ValidateDiagFunc` because `ValidateFunc` is [deprecated](https://www.terraform.io/plugin/sdkv2/guides/v2-upgrade-guide#deprecation-of-helper-schema-schemavalidatefunc) in SDK v2.

ka-myl: Funny. I don't remember when was the last time I checked the docs for `ValidateFunc`, but I'm pretty sure the deprecation notice wasn't there yet. Anyway, in this case I think I'll just close this PR, as the `validation` package doesn't seem to be adjusted to that change yet. We could do some sort of wrapper that would transform `SchemaValidateFunc` into `SchemaValidateDiagFunc`, but imo thats just producing more code that will be removed at some point in the future

peknur: There is [ToDiagFunc](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/helper/validation/meta.go#L65) function that we could probably use?

Something like:
`ValidateDiagFunc: validation.ToDiagFunc(validation.StringMatch(
	regexp.MustCompile(`^[a-z]{2}_[A-Z]{2}\.[A-z0-9-]+$`),
	"invalid locale; must be in form en_US.UTF8 (language_TERRITORY.CODEPOINT)"))

Other that that this PR looks good to me and no need to close it :)

peknur (approved):
Looks good to me. Nicely simplified.

PR #216, opened 2022-01-17
Add release dates to changelog entries

vtorhonen:
Currently changelog entries are missing release dates. This PR adds release dates of previous releases. Uses ISO 8601 date format.


ka-myl (approved):
LGTM

PR #217, opened 2022-01-17
fix(server): subaccount tagging error handling

peknur:
Return more descriptive error message if subaccount tries to
edit server tags.

Update Go API dependency from v4.1.2 to v4.1.3 to get access to
new account functions.


ka-myl (approved):
LGTM

PR #221, opened 2022-01-21
Fix build-time variable injection for release builds

vtorhonen:
Build-time variable injection was modified on #164 and #165 where the variable was moved from `globals` package to `config`. Since then the release pipeline has still been referencing the old package. Due to this configuration issue the version information has not made it into the release builds. API client uses this information for initialising the HTTP user agent correctly.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -16,7 +16,7 @@ builds:
     - -trimpath
   ldflags:
     - -s -w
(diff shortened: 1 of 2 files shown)


peknur (approved):
Good catch! I'm not that familiar with GoReleaser but looks good to me.

Darep (approved):
LGTM! If it fixes it then it fixes it 😄  Let's release this soon 🙏

PR #222, opened 2022-01-25
fix(server): missing template id

peknur:
Fix missing template id if resource creation fails on tag errors.
Missing id will prevent deleting the resource.


Darep (approved):
The code change looks good to me! The storage UUID should be in the original "create server"-payload already so should work 👍

ka-myl (approved):

PR #226, opened 2022-01-28
feat(storage): upcloud_storage data source

peknur:
New upcloud_storage data source to retrieve specific storage details.


upcloud/datasource_upcloud_storage.go
```diff
@@ -0,0 +1,172 @@
...
+		return diag.Errorf("query returned no results")
+	}
+
+	if len(matches) > 1 {
+		if !d.Get("most_recent").(bool) {
+			return diag.Errorf("query returned more than one result")
+		}
+		// sort storages by created timestamp
+		sort.Slice(matches, func(i, j int) bool {
+			return matches[i].Created.Unix() > matches[j].Created.Unix()
```
ka-myl, line 142: This will not really work as expected I think. I don't know why `go-api-sdk` includes `Created` field for storage, but it does not seem to be populated (and it looks like it's intentional). So all storages will have the same created time, and that will be the default value to `time.Time{}`.

You can see this with this config for example:
```

data "upcloud_storage" "ubuntu_template" {
  type = "template"
  name_regex = "^Ubuntu"
  most_recent = true
}

resource "upcloud_server" "example" {
  hostname = "ubuntu-ecample"
  zone     = "pl-waw1"
  plan     = "1xCPU-1GB"

  template {
    storage = data.upcloud_storage.ubuntu_template.id
    size    = 25

    backup_rule {
      interval  = "daily"
      time      = "0100"
      retention = 8
    }
  }

  network_interface {
    type = "public"
  }

  network_interface {
    type = "utility"
  }
}
```

The server will be created with Ubuntu 18 template, even tho we have Ubuntu 20 available

peknur: Thanks @ka-myl ! I'll see what is happening here. With custom images it seems to be working but I'll test it also to be sure.

ka-myl (approved):
LGTM

PR #228, opened 2022-02-08
fix(provider): handle dangling resource references

peknur:
Remove a binding to an remote object if it no longer exists.
This is to some degree equivalent to 'terraform state rm' command.


upcloud/provider.go
```diff
@@ -132,3 +142,17 @@ func isProviderAccountSubaccount(s *service.Service) (bool, error) {
...
+		Summary:  err.ErrorMessage,
+		Detail:   details,
+	}
+}
+
+func diagBindingRemovedWarningFromUpcloudErr(err *upcloud.Error, name string) diag.Diagnostic {
+	return diagWarningFromUpcloudErr(err,
+		fmt.Sprintf("Binding to an existing remote object '%s' will be removed from the state. Next plan will include action to re-create the object if you choose to keep it in config.", name),
+	)
+}
```
ka-myl, line 158: Would it be worthwhile to have a separate package for those error constants and functions? I find it really confusing when we use them in another files where they are neither imported nor declared. It's a bit like having global variables for me.
Or is it some sort of go convention?

ka-myl (approved):
LGTM. Left one comment but it's not a blocker

upcloud/provider.go
peknur: @ka-myl thanks for review! Everything inside that folder belongs to `upcloud` package so basically you can put functions/variables to any file you want and it will be "visible" inside that upcloud directory without importing file. You could  combine every Go file in that directory to one e.g. `uplcoud.go` file and it would work same way.  I chose to put them in `provider.go` file because these are kind of provider (UpCloud) related functions/constants and these are referenced in almost all of the `resource_upcloud_*.go` files. So this is partly Go convention and partly my interpretation of what could be a good way to do it :)

That being said we probably should start slicing this package to smaller pieces at some point to make it more readable.

PR #229, opened 2022-02-09
Fix/importing managed db

ka-myl:
Fix a runtime error when importing managed db.

The error was caused by the fact that `type` is set by provider during creation. So when we import the db, `d.Get("type")` does not actually return anything, which ten causes some problems down the line


peknur (approved):
LGTM

PR #233, opened 2022-02-16
Feat/objsto keys from env vars

ka-myl:
This PR:
- allows using env variables for object storage access and secret key
- enables object storage import

### DIFF
--- a/upcloud/resource_upcloud_objectstorage.go
+++ b/upcloud/resource_upcloud_objectstorage.go
@@ -440,7 +440,7 @@ func modifyObjectStorage(client *service.Service, req *request.ModifyObjectStora
 }
 
 // Attempts to get access key.
-// Second return value is a bool, set to true if key value was retrived from env variable
+// Second return value is a bool, set to true if key value was retrieved from env variable
 func getAccessKey(d *schema.ResourceData) (string, bool, error) {
 	configVal := d.Get("access_key").(string)
 
--- a/upcloud/resource_upcloud_objectstorage.go
+++ b/upcloud/resource_upcloud_objectstorage.go
@@ -3,7 +3,6 @@ package upcloud
 import (
 	"context"
 	"fmt"
-	"log"
 	"net/url"
 	"os"
 	"strings"
@@ -37,6 +36,9 @@ func resourceUpCloudObjectStorage() *schema.Resource {
 		ReadContext:   resourceObjectStorageRead,
 		UpdateContext: resourceObjectStorageUpdate,
 		DeleteContext: resourceObjectStorageDelete,
+		Importer: &schema.ResourceImporter{
+			StateContext: schema.ImportStatePassthroughContext,
+		},
 		Schema: map[string]*schema.Schema{
 			"size": {
 				Description:  "The size of the object storage instance in gigabytes",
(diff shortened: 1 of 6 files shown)


upcloud/resource_upcloud_objectstorage.go
```diff
@@ -380,3 +438,117 @@ func modifyObjectStorage(client *service.Service, req *request.ModifyObjectStora
...
+	name := strings.ToUpper(strings.Replace(objectStorageName, "-", "_", -1))
+	return fmt.Sprintf("%s%s", prefix, name)
+}
+
+type objectStorageKeyType string
+
+const (
+	objectStorageKeyTypeAccess objectStorageKeyType = "access_key"
+	objectStorageKeyTypeSecret objectStorageKeyType = "secret_key"
+)
```
peknur, line 516: Should we move these package constants inside `createKeyValidationFunc`  for now to make them local only ?

upcloud/resource_upcloud_objectstorage.go
```diff
@@ -380,3 +438,117 @@ func modifyObjectStorage(client *service.Service, req *request.ModifyObjectStora
...
+
+func createKeyValidationFunc(attrName objectStorageKeyType, minLength, maxLength int) schema.SchemaValidateDiagFunc {
+	return func(val interface{}, path cty.Path) diag.Diagnostics {
+		key := val.(string)
```
peknur, line 520: This is not big one but we could check if type assertion fails:
```
key, ok := val.(string)
if !ok {
   return diag.FromErr(fmt.Errorf("expected type of %s to be string", val))
}
```

upcloud/resource_upcloud_objectstorage.go
```diff
@@ -17,6 +22,12 @@ import (
 
 const bucketKey = "bucket"
 const numRetries = 5
+const AccessKeyEnvVarPrefix = "UPCLOUD_OBJECT_STORAGE_ACCESS_KEY_"
+const SecretKeyEnvVarPrefix = "UPCLOUD_OBJECT_STORAGE_SECRET_KEY_"
```
peknur, line 26: I think that exporting these at the moment is not necessary, Our tests lives also inside `upcloud` package.

upcloud/resource_upcloud_objectstorage.go
```diff
@@ -380,3 +438,117 @@ func modifyObjectStorage(client *service.Service, req *request.ModifyObjectStora
...
+			case objectStorageKeyTypeAccess:
+				envVarPrefix = AccessKeyEnvVarPrefix
+			case objectStorageKeyTypeSecret:
+				envVarPrefix = SecretKeyEnvVarPrefix
+			default:
+				return diag.Errorf("unknown attribute name for creating object storage keys validation function: %s; this is a provider error", attrName)
+			}
+
+			if !utils.EnvKeyExists(envVarPrefix) {
+				return diag.Errorf("%s set to empty string, but no environment variables for it found", attrName)
```
peknur, line 536: Should we add also hint about variable name to that error message  e.g.
` diag.Errorf("%s set to empty string, but no environment variables for it found (%s{NAME})", attrName, envVarPrefix)`

peknur (approved):
Good job! Looks good to me. Tested this locally and importing also works now nicely. Left some comments to fine tune code but nothing super critical

upcloud/resource_upcloud_objectstorage.go
ka-myl: It technically should not be possible, since the type is defined in schema. Then again the runtime type assertion errors do happen from time to time in development, so I guess it would be good to get into habit of checking those, even if terraform sdk guarantees them :P

PR #234, opened 2022-02-18
Feat/object storage import temp creds

ka-myl:
This PR adds an option to pass session token in env vars when importing object storage.

### DIFF
--- a/upcloud/resource_upcloud_objectstorage.go
+++ b/upcloud/resource_upcloud_objectstorage.go
@@ -24,6 +24,7 @@ const bucketKey = "bucket"
 const numRetries = 5
 const accessKeyEnvVarPrefix = "UPCLOUD_OBJECT_STORAGE_ACCESS_KEY_"
 const secretKeyEnvVarPrefix = "UPCLOUD_OBJECT_STORAGE_SECRET_KEY_"
+const tokenEnvVarPrefix = "UPCLOUD_OBJECT_STORAGE_TOKEN_"
 const accessKeyMinLength = 4
 const accessKeyMaxLength = 255
 const secretKeyMinLength = 8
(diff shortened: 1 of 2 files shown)


upcloud/resource_upcloud_objectstorage.go
```diff
@@ -313,7 +315,18 @@ func copyObjectStorageDetails(objectDetails *upcloud.ObjectStorageDetails, d *sc
 		_ = d.Set("secret_key", "")
 	}
 
-	buckets, err := getBuckets(objectDetails.URL, accessKey, secretKey)
+	// This is an undocumented feature, can be changed or removed at any time without any announcements
+	// It allows you to also specify session token in env variables, which will be then passed to minio client
+	// along with access and secret key. This allows you to use temp credentials to read buckets list and such.
+	// This is only intended for import.
+	token := getToken(d)
```
ka-myl, line 322: I kind of dislike that this checking for token and passing it to `getBucketConnection` is so implicit. It is very concise, but if the token is not set, if just passes empty string to `getBucketConnection` and there is no indication in the code that the lack of token was handled.
At the beginning I wanted to make those two things very separate in the code, like if there is a token set in env vars, we have a completely different code path. But that would duplicate a lot of code, and also gets a bit complicated with the fact that you totally can have secret key in env var, but access key in the config.
So at the end of the day I decided this ok-ish. It is implicit, but it's also very easy to remove if we ever want to do that

peknur: What if we would change `getBucketConnection` signature to get `*upcloud.ObjectStorageDetails` as first argument e.g.
```
func getBucketConnection(objsto *upcloud.ObjectStorageDetails, accessKey, secretKey  string)
```
And them move that `getToken` inside that function and change it to `token := getToken(objsto.Name)`

ka-myl:
It seems like this will not be needed at the end of the day, closing the PR

PR #235, opened 2022-03-03
Feat/autoresize parition and filesystem

ka-myl:
This PR adds an autoresize partition and fs feature for storages and server template storages


peknur (approved):
Tested this locally and didn't find any problems. Good job!

PR #236, opened 2022-03-04
fix(dbaas): fix database properties

peknur:
- Fix PostgreSQL properties that has dot: pg_stat_statements_track, pg_partman_bgw_role, pg_partman_bgw_interval
- Move schema definitions from JSON to Go to allow more advanced field properties


ka-myl (approved):
Code looks good. Tested locally and also seems to be working as intended. Nice work here :D

PR #237, opened 2022-03-07
refactor(storage): combine multiple acceptance tests into one

ka-myl:
This PR just combines a few storage acceptance tests into one. Those tests actually just tested different schema properties and updated to them, so it's better to combine them into one to save some time.

I also removed the function that creates config string and replaced it with actual string. While this might not be fully dry, IMO it's easier to just eyball the config as you write the checks for said config. Having to scroll few hundreds line down to check what is in the actual config is imo annoying


peknur (approved):
I completely agree that embedding those config strings makes it easier to read. Good job!

PR #245, opened 2022-03-31
Fix/resolvers required fields

ka-myl:
Mark resolvers `nameservers` field as required.
Move domain name (or hostname) validation to separate package
Add validation for all resources names
Make sure different tests use private networks with different ip ranges


peknur (approved):
Nice! Looks good!

PR #261, opened 2022-05-05
fix(dbaas): maintenance window properties

peknur:
- require that both `maintenance_window_time` and `maintenance_window_dow` are set when defining maintenance window
- fix `maintenance_window_time` format


ka-myl (approved):
LGTM

PR #267, opened 2022-05-10
chore: update changelog for v2.4.2

peknur:


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+## [2.4.2] - 2022-05-10
+
 ### Changed
 - Update GoReleaser to v1.8.3
```
Darep, line 11: I would also add here something like `- Re-added missing binary for Darwin arm64` (though it might not be correct in case the build fails at it again 😅 )

Darep (approved):

ka-myl (approved):

CHANGELOG.md
peknur: Yep :) I thought that we could mention that in release notes if this works

PR #273, opened 2022-06-09
Update upcloud-go-api version and add context where relevant

ka-myl:
This PR updates upcloud-go-api package to the version that adds context-aware client and service, and updates the required method calls.


peknur (approved):
Tested this locally and didn't find any problems. Looks good!

PR #283, opened 2022-08-02
Adds password delivery methods to documentation

jaamo:
I had to check source code for possible values for password_delivery property. These could be added to documentation.


kangasta (approved):

kangasta:
Thank you for the pull request!

I copied the change also to related string in source code, so that the change wont disappear when docs are next auto-generated, and updated the commit message to follow [convetional commits](https://www.conventionalcommits.org/en/v1.0.0/).

PR #285, opened 2022-08-08
Add action for auto-generating docs

kangasta:
- [x] Depends on #284


kangasta:
Created a new PR (#287) from `chore/docs-generation` for this to allow creating `main` braches branch protection rules for `test-docs-generator` for testing the auto commit step.

PR #287, opened 2022-08-10
Add action for auto-generating docs

kangasta:
Testing results (1st commit):

- See #289 for test PR that adds extra text to `docs/` that will be removed as it is not found from templates or source-code.
- See #290 for PR created by this new action. (Edited PR body manually to test if GitHub changes Commit ID automatically into a link)

Testing result (2nd commit):

- #292
- #293

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,39 @@
+name: Documentation
+
+on:
+  push:
+    branches:
+      - main
+      - test-docs-generator # for testing
+
+jobs:
+  update:
+    name: Update
+    runs-on: ubuntu-latest
+    steps:
+    - name: Setup Go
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Generate documentation
+      run: |
+        terraform version
+        make docs
+    - name: Create PR for docs update
+      uses: peter-evans/create-pull-request@v4
+      with:
+        add-paths: docs/
+        branch: chore/update-docs
+        commit-message: "chore(docs): update documentation"
```
villevsv-upcloud, line 35: Could `commit-message` refer the changes that triggered the action, pretty much like in the `body`?

villevsv-upcloud (approved):

PR #294, opened 2022-08-31
fix: set credentials as optional inputs

kangasta:
This does not change how these inputs are used: providing these values in the provider block has already been optional if credentials were defined as environment variables.


kangasta:
Acceptance tests fail probably because DB properties naming change in API. Should be fixed by #295

villevsv-upcloud (approved):

PR #295, opened 2022-08-31
fix(dbaas): remove special handling of dots in postgres properties

kangasta:
As of DBaaS 1.2.0 all properties are in snake case.


peknur (approved):
Looks good!

PR #297, opened 2022-09-06
feat(storage): update maximum storage size from 2048 to 4096 gigabytes

peknur:


kangasta (approved):

kangasta:
I saw the same `Check 10/13 error: upcloud_managed_database_postgresql.pg2: Attribute 'properties.0.version' expected "14", got "13"` error when working with #295. The tests seemed to randomly pass/fail as the 5 min wait for DB to be fully running before upgrading version seems not be enough. Should we increase the timeout [here](https://github.com/UpCloudLtd/terraform-provider-upcloud/blob/main/internal/service/database/database.go#L175)?

peknur:
I think that I've seen that error also when some other part of test is failing which I think was also the problem this time.  But yeah we can increase that timeout value also. Those DB operations sometimes take long time.

Issue #182, opened 2021-10-03
What's the correct way to get list of Operating System


sentabi:
Right now I use 
```
storage = "Ubuntu Server 20.04 LTS (Focal Fossa)"
```
or 
```
storage = "01000000-0000-4000-8000-000030200200"
```
Which one is the recommended approach, the [official blog using UUID](https://upcloud.com/community/tutorials/automate-terraform-deployments/), while the terraform [example](https://github.com/UpCloudLtd/terraform-provider-upcloud/tree/master/examples) using Distro Name

and how do I get the list of supported operating systems?

My use case, I need to test the script on a different version OS (ex Debian 10, Debian 11, CentOS 7, and CentOS 8)

Darep:
Both should work nicely, I don't think there is a "correct" way 😊  I would personally use the name as it will "self-document" what template you are using. With the UUID, you need to add a comment so you know what template the UUID points to.

> and how do I get the list of supported operating systems?

Probably the best way at the moment is to log in to https://hub.upcloud.com and go to Storage > Templates and see the names & UUIDs there.

PopoSensei:
You can also query that info straight from UpCloud API https://developers.upcloud.com/1.3/9-storages/  `GET /1.3/storage/template`

sentabi:
thanks @PopoSensei @Darep

PR #321, opened 2022-11-09
feat(lbaas): private network support

peknur:
Adds support for load balancers listening private networks.  Built on top of [upcloud-go-api](https://github.com/UpCloudLtd/upcloud-go-api) PR [#182](https://github.com/UpCloudLtd/upcloud-go-api/pull/182)

Bumped Go version to 1.8 and Terraform SDK to v2.24.0.


nate-droid (approved):

examples/resources/upcloud_loadbalancer/resource.tf
```diff
@@ -19,4 +19,15 @@ resource "upcloud_loadbalancer" "lb" {
   plan              = "development"
   zone              = var.lb_zone
   network           = resource.upcloud_network.lb_network.id
```
ka-myl, line 21: This line should probably be removed from the example

internal/service/loadbalancer/loadbalancer.go
```diff
@@ -42,11 +45,16 @@ func ResourceLoadBalancer() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 			},
-			"network": {
-				Description: "Private network UUID where traffic will be routed. Must reside in load balancer zone.",
-				Type:        schema.TypeString,
-				ForceNew:    true,
-				Required:    true,
+			"networks": {
```
ka-myl, line 48: I would use a singular form here, i.e. `network`. Singular form is a bit nicer from a user perspective, for me having to use a plural form to define a single network is a bit confusing. Not a blocker

internal/service/loadbalancer/loadbalancer.go
```diff
@@ -89,17 +97,144 @@ func ResourceLoadBalancer() *schema.Resource {
...
+		"network": {
+			Description: "Private network UUID. Required for private networks and must reside in loadbalancer zone. For public network the field should be omitted.",
+			Type:        schema.TypeString,
+			ForceNew:    true,
+			Optional:    true,
+		},
+		"dns_name": {
+			Description: "DNS name of the load balancer network",
+			Type:        schema.TypeString,
+			ForceNew:    true,
```
ka-myl, line 171: Probably not needed, as this is a computed only attribute

ka-myl (approved):
LGTM. Tested locally and it works nicely. Left some minor comments, but those are just points to consider, no blockers!

internal/service/loadbalancer/loadbalancer.go
peknur: good catch, I'll fix that.

internal/service/loadbalancer/loadbalancer.go
peknur: That is true. I was thinking to use singular form but examples that I viewed all used plural naming and `network` is still reserved to old field so I decided to go with plural.

examples/resources/upcloud_loadbalancer/resource.tf
peknur: Yes it should :) It's removed now.

PR #339, opened 2023-01-11
refactor(kubernetes): handle node groups as separate resource

peknur:
- handle node groups as separate resource
- change node group to accept list of kubelet arguments
- rename node group `count` (reserved name by TF) field to `node_count`
- move resource error handling under `utils` package from individual
  packages


nate-droid (approved):

internal/service/kubernetes/node_group.go
```diff
@@ -0,0 +1,296 @@
...
+			},
+			"name": {
+				Description:      "The name of the node group. Needs to be unique within a cluster.",
+				Type:             schema.TypeString,
+				Required:         true,
+				ValidateDiagFunc: validateResourceName,
+				ForceNew:         true,
+			},
+			"plan": {
+				Description: "The pricing plan used for the node group. Valid values available via `upcloud_kubernetes_plan` datasource field `description`.",
```
kangasta, line 47: Now that we do not have separate `K8S-` plans, this could mention that any server plan can be used and those can be listed with `upctl server plans`. This could be updated in later PR as well.

villevsv-upcloud (approved):

internal/service/kubernetes/node_group.go
peknur: PR [#340](https://github.com/UpCloudLtd/terraform-provider-upcloud/pull/340)

PR #348, opened 2023-01-30
Feat/k8s aa node groups

ka-myl:


internal/service/kubernetes/node_group.go
```diff
@@ -49,6 +49,14 @@ func ResourceNodeGroup() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 			},
+			"anti_affinity": {
+				Description: `If set to true, nodes in this group will be placed on separate compute hosts.
+				Please note that anti-affinity policy is not fully guaranteed that the nodes will end up on different hardware.`,
```
kangasta, line 54: `... anti-affinity policy does not fully guarantee that ...` would sound a bit more fluent to me.

kangasta (approved):

internal/service/kubernetes/node_group.go
ka-myl: Well, it would be grammatically correct, as opposed to the current one 🤦

PR #351, opened 2023-02-20
fix(kubernetes): provide data source certs and keys in PEM format

kangasta:


villevsv-upcloud (approved):

kangasta:
Test failure caused by slow responses from AU-SYD1, those are unrelated to this change.

PR #376, opened 2023-05-25
feat(dbaas): OpenSearch service type support

villevsv-upcloud:


internal/service/database/database.go
```diff
@@ -375,8 +351,28 @@ func waitServiceNameToPropagate(ctx context.Context, name string) (err error) {
 	return errors.New("max retries reached while waiting for service name to propagate")
 }
 
-func updateDatabaseVersion(ctx context.Context, d *schema.ResourceData, client *service.Service) diag.Diagnostics {
-	diags := diag.Diagnostics{}
+func updateDatabaseVersion(ctx context.Context, d *schema.ResourceData, client *service.Service) (diags diag.Diagnostics) {
```
peknur, line 354: I seems that this is now only referenced by PostgreSQL resource, was intentional ?

internal/service/database/postgresql.go
```diff
@@ -27,6 +30,72 @@ func ResourcePostgreSQL() *schema.Resource {
...
+
+func resourcePostgreSQLUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	diags := resourceDatabaseUpdate(ctx, d, meta)
+	if diags.HasError() {
+		return diags
+	}
+
+	client := meta.(*service.Service)
+
+	switch d.HasChange("powered") {
```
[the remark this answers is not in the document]
villevsv-upcloud: True! Replaced the top level switch and moved the shorter block to be first.

internal/service/database/database.go
villevsv-upcloud: It's intentional, as the upgrade operation is only allowed for PostgreSQL resources for now. I opted to use `ForceNew` if user wants to update the version of and OpenSearch resource.

peknur: There is still `version` property in MySQL resource, should we still handle that also ?

villevsv-upcloud: Added `ForceNew: true` to MySQL as well. Currently there is only one whitelisted version (8), so the update issue has not arisen (yet).

thevilledev:
Requires #379 to be merged in first

peknur (approved):
Looks good :+1:

PR #378, opened 2023-05-29
feat(kubernetes): private node groups support

peknur:


internal/service/kubernetes/kubernetes.go
```diff
@@ -71,6 +71,13 @@ func ResourceCluster() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 			},
+			"private_node_groups": {
+				Description: "Enable private node groups. Private node groups requires network that is routed through NAT gateway.",
```
villevsv-upcloud, line 75:
```suggestion
				Description: "Enable private node groups. Private node groups require a network that is routed through NAT gateway.",
```

villevsv-upcloud (approved):
LGTM

internal/service/kubernetes/kubernetes.go
peknur: Good catch :+1:  Thanks @villevsv-upcloud for the review.

PR #401, opened 2023-09-04
feat(kubernetes): control plane ip filter for cluster resource

villevsv-upcloud:


examples/resources/upcloud_kubernetes_cluster/resource.tf
```diff
@@ -17,9 +17,11 @@ resource "upcloud_network" "example" {
 
 # Create a Kubernetes cluster
 resource "upcloud_kubernetes_cluster" "example" {
-  name    = "exampleapp"
-  network = upcloud_network.example.id
-  zone    = "de-fra1"
+  # Allow access the cluster control plane from any external source.
```
kangasta, line 20:
```suggestion
  # Allow access to the cluster control plane from any external source.
```

upcloud/testdata/upcloud_kubernetes/kubernetes_s2.tf
```diff
@@ -23,17 +23,18 @@ resource "upcloud_network" "main" {
 }
 
 resource "upcloud_kubernetes_cluster" "main" {
-  name    = var.name
-  network = upcloud_network.main.id
-  zone    = var.zone
+  control_plane_ip_filter = ["0.0.0.0/0"]
```
kangasta, line 26: Could also test changing this value similarly than node-group node count change is now tested.

kangasta (approved):
Added few minor comments, otherwise LGTM 👌

upcloud/testdata/upcloud_kubernetes/kubernetes_s2.tf
villevsv-upcloud: Thanks, added a check for empty set.

PR #404, opened 2023-09-28
feat(router): static routes support

villevsv-upcloud:


upcloud/testdata/upcloud_gateway/gateway_s1.tf
```diff
@@ -10,6 +10,10 @@ variable "zone" {
 
 resource "upcloud_router" "this" {
   name = "${var.prefix}router"
+
+  lifecycle {
+    ignore_changes = [static_route]
```
kangasta, line 15: Does the gateway automatically set `static_routes` to the router? Could we somehow handle that automatic route so that end-user does not have to use `lifecycle.ignore_changes` 🤔 Or is it something we could instruct user to do themselves, like the router creation with Kubernetes clusters.

villevsv-upcloud: Added a comment to the documentation for now.

kangasta: The static routes automatically set by the gateway seem to be in link-local address space (`169.254.0.0/16`). Thus, they should not collide with values expected from user. Maybe some sort special value would work here, but that would require more investigating/testing.

kangasta: Probably good to merge this like this to unblock the release and then try to add fix or workaround for this later 🤔

kangasta (approved):

PR #428, opened 2023-10-25
fix(kubernetes): node_group re-creation waits for deletion first

villevsv-upcloud:
TODO:

- [x] upcloud-go-api [v6.8.3](https://github.com/UpCloudLtd/upcloud-go-api/pull/273)


peknur (approved):
Looks good :+1:  Tested it locally and everything seems to works as expected. 
I have a bit mixed feelings on should we wait node group(s) state when creating the service, but I guess it's always better to have service fully running at the end and this will also remove node group state update after creation.

PR #438, opened 2023-11-03
fix(dbaas): wait master node to enter running state

peknur:


villevsv-upcloud (approved):
Looks good!

peknur:
Closing this PR, because it seems that problem was actually caused by temporary problem with API.

PR #443, opened 2023-11-16
docs: group resources and data-sources by product

kangasta:


docs/data-sources/hosts.md
```diff
@@ -1,7 +1,7 @@
 ---
 # generated by https://github.com/hashicorp/terraform-plugin-docs
 page_title: "upcloud_hosts Data Source - terraform-provider-upcloud"
-subcategory: ""
+subcategory: Cloud
```
kangasta, line 4: Cloud category would include these non-product resources and data-sources. This could also be `Other`.

villevsv-upcloud: Cloud seems reasonable to me.

villevsv-upcloud (approved):

PR #515, opened 2024-03-14
Feat/vpnaas

ka-myl:
This PR adds support for VPNaaS


internal/service/gateway/tunnel.go
```diff
@@ -0,0 +1,496 @@
...
+				Description: "Remote public IP address of the tunnel",
+				Type:        schema.TypeString,
+				Required:    true,
+			},
+			"operational_state": {
+				Description: "Tunnel's current operational, effective state",
+				Type:        schema.TypeString,
+				Computed:    true,
+			},
+			"ipsec_auth_psk": {
```
ka-myl, line 80: This is an artificial field in a sense that in API, it is actually nested in `ipsec` block. For example, a POST request to create a tunnel would look like this:
```
{
  "name": "example-tunnel-2",
  "local_address": {
    "name": "public-ip-1"
  },
  "remote_address": {
    "address": "100.10.0.111"
  },
  "ipsec": {
    "authentication": {
        "authentication": "psk",
        "psk": "{your PSK}"
    }
  }
}

```
I have extracted the authentication part to a separate schema block to avoid weird validation issues when we start adding other authentication types. If we need to support a new auth type, we can just implement it as a separate block with it's own set of properties and use `ExactlyOneOf` schema behaviour to make sure user defines only one. This should give us a little more flexibility in the future (I hope :P)

internal/service/gateway/gateway.go
```diff
@@ -92,7 +95,47 @@ func ResourceGateway() *schema.Resource {
...
+				Description: planDescription,
+				Computed:    true,
+				Optional:    true,
+				Type:        schema.TypeString,
+			},
+			"address": {
+				Description: addressesDescription,
+				Computed:    true,
+				Optional:    true,
+				Type:        schema.TypeSet,
```
villevsv-upcloud, line 108: Could this be `schema.TypeMap`?

villevsv-upcloud: Oops, sorry. It could not. :)

villevsv-upcloud: Is the max limit of 1 about to change at some point? If not, we could have `ip_address` and `ip_address_name` fields instead of the `address`

ka-myl: It's quite possible it will change at some point :)

ka-myl:
Unrelated tests failed (UKS & DBaas) with something that seems like a known issues

upcloud/resource_upcloud_gateway_test.go
```diff
@@ -26,21 +30,116 @@ func TestAccUpcloudGateway(t *testing.T) {
...
+					resource.TestCheckResourceAttr(conn2Name, "remote_route.0.name", "remote-route2"),
+					resource.TestCheckResourceAttr(conn2Name, "remote_route.0.type", "static"),
+					resource.TestCheckResourceAttr(conn2Name, "remote_route.0.static_network", "222.123.123.0/24"),
+
+					resource.TestCheckResourceAttr(tunnel2Name, "name", "test-tunnel2"), // Single check for tunnel2 just to make sure it gets created
+
+					// This field is deprecated, can be removed later
+					resource.TestCheckTypeSetElemNestedAttrs(name, "addresses.*", map[string]string{"name": "my-public-ip"}),
+				),
+			},
```
villevsv-upcloud, line 91: Adding a new step with something like below would test the import as well:
```
			ResourceName:      name,
			ImportState:       true,
			ImportStateVerify: true,
```

villevsv-upcloud (approved):
One `ImportState` mention but looks good

upcloud/resource_upcloud_gateway_test.go
ka-myl: Good catch, done

PR #522, opened 2024-04-11
chore(go): bump to version 1.21

villevsv-upcloud:


.github/workflows/docs.yml
```diff
@@ -15,21 +15,21 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
ka-myl, line 18: Why commit hash instead of just `v4.1.1`?

ka-myl (approved):

.github/workflows/docs.yml
villevsv-upcloud: Using hash was originally introduced in some of other our repositories, it covers the extremely rare case that the tag has been re-used a later point in time.

PR #523, opened 2024-04-12
Fix/objsto read error handling

ka-myl:
This PR fixes the issue for all managed object storage resources, when apply/refresh errors out when you create a resource with TF and then delete it outside TF. The issue was that we didn't remove resource ID from the state on `not found` errors in read functions.


villevsv-upcloud (approved):

peknur (approved):
Looks good :+1

PR #543, opened 2024-05-21
feat(gateway): add UUID support for connections and tunnels

villevsv-upcloud:


internal/service/gateway/connection.go
```diff
@@ -72,6 +138,24 @@ func ResourceConnection() *schema.Resource {
...
+			rawState["uuid"] = conn.UUID
+
+			return rawState, nil
+		}
+	}
+
+	return rawState, fmt.Errorf("connection by name %s not found", rawState["name"].(string))
+}
+
 func resourceConnectionCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
```
kangasta, line 159: Should we use the UUID of the created resource in the `d.SetId(...)`? (Could not add comment to unchanged line 176)

internal/service/gateway/tunnel.go
```diff
@@ -94,14 +99,106 @@ func ResourceTunnel() *schema.Resource {
...
+			rawState["uuid"] = tunnel.UUID
+
+			return rawState, nil
+		}
+	}
+
+	return rawState, fmt.Errorf("tunnel by name %s not found", name)
+}
+
 func resourceTunnelCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) (diags diag.Diagnostics) {
```
kangasta, line 196: Here also name is used in the id instead of UUID 🤔 Is changing the id possible in the upgrader?

internal/service/gateway/connection.go
villevsv-upcloud: Changed to use the UUID, better for import purposes.

internal/service/gateway/tunnel.go
villevsv-upcloud: Changed to use the UUID, better for import purposes.

internal/service/gateway/connection.go
```diff
@@ -69,9 +75,88 @@ func ResourceConnection() *schema.Resource {
...
+			},
+			"remote_route": {
+				Description:  "Route for the remote side of the network.",
+				Type:         schema.TypeSet,
+				Optional:     true,
+				AtLeastOneOf: []string{"local_route", "remote_route"},
+				Elem:         gatewayRouteSchema(),
+			},
+			"tunnels": {
+				Description: "List of connection's tunnels names. Note that this field can have outdated information as connections are created by a separate resource. To make sure that you have the most recent data run 'terrafrom refresh'.",
```
kangasta, line 130: Does this now take in UUIDs
```suggestion
				Description: "List of connection's tunnels UUIDs. Note that this field can have outdated information as connections are created by a separate resource. To make sure that you have the most recent data run 'terrafrom refresh'.",
```

kangasta (approved):

internal/service/gateway/connection.go
villevsv-upcloud: Still names. We could revise this in further updates.

PR #551, opened 2024-05-29
test(gateway): add test case that omits all optional fields

kangasta:
At least one of the data-consistency errors (default value for `configured_status`) requires re-implementation with plugin framework, see also [Hashicorp docs](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/data-consistency-errors#planned-value-for-a-non-computed-attribute).

### DIFF
--- a/upcloud/gateway/resource_upcloud_gateway_test.go
+++ b/upcloud/gateway/resource_upcloud_gateway_test.go
@@ -164,6 +164,36 @@ func TestAccUpcloudGateway(t *testing.T) {
 	})
 }
 
+func TestAccUpcloudGateway_Minimal(t *testing.T) {
+	testData := utils.ReadTestDataFile(t, "testdata/gateway_minimal.tf")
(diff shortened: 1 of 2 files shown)


internal/service/gateway/gateway.go
```diff
@@ -34,11 +34,13 @@ const (
 
 func ResourceGateway() *schema.Resource {
 	return &schema.Resource{
-		Description:   "Network gateways connect SDN Private Networks to external IP networks.",
-		CreateContext: resourceGatewayCreate,
-		ReadContext:   resourceGatewayRead,
-		UpdateContext: resourceGatewayUpdate,
-		DeleteContext: resourceGatewayDelete,
+		EnableLegacyTypeSystemApplyErrors: true,
```
kangasta, line 37: Probably better to remove the commit that adds these `EnableLegacyTypeSystem*Errors` for now and look into re-implementing this resource with plugin framework 🤔

PR #556, opened 2024-06-06
feat(router): add `labels` support, re-implement with plugin framework

kangasta:
Also fixes error handling in `Read` functions of `upcloud_network` and `upcloud_network_peering` so that these resources are removed from state, if API returns 404 response.


villevsv-upcloud (approved):
Looks good!

While at it, maybe adding an example with a lifecycle block & `ignore_changes = [static_route]` would be nice, as it comes up every now and then. We seem to already have it in the gateway examples.

PR #564, opened 2024-06-25
refactor(kubernetes): migrate cluster data-source to plugin framework

kangasta:


internal/service/kubernetes/cluster_data_source.go
```diff
@@ -0,0 +1,161 @@
...
+			},
+			"id": schema.StringAttribute{
+				Required:    true,
+				Description: idDescription,
+			},
+			"host": schema.StringAttribute{
+				Computed:    true,
+				Description: hostDescription,
+			},
+			"kubeconfig": schema.StringAttribute{
```
kangasta, line 91: Should `kubeconfig` also be sensitive as it includes the `client_key` value? 🤔

villevsv-upcloud: Makes sense to mark it as sensitive

villevsv-upcloud (approved):

PR #572, opened 2024-07-01
refactor(cloud): migrate hosts and zones data source to plugin framework

villevsv-upcloud:


go.mod
```diff
@@ -7,12 +7,12 @@ require (
 	github.com/hashicorp/go-cty v1.4.1-0.20200414143053-d3edf31b6320
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/hashicorp/go-uuid v1.0.3
-	github.com/hashicorp/terraform-plugin-framework v1.7.0
+	github.com/hashicorp/terraform-plugin-framework v1.4.1
```
kangasta, line 10: Are these downgrades included here on purpose or leftovers from #571?

internal/service/cloud/zones_data_source.go
```diff
@@ -0,0 +1,108 @@
...
+	zones, err := d.client.GetZones(ctx)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Unable to read zones",
+			utils.ErrorDiagnosticDetail(err),
+		)
+		return
+	}
+
+	data.ID = types.StringValue(time.Now().UTC().String())
```
kangasta, line 89: Would this work with `"zones" + data.FilterType` as the ID 🤔 There should not be need for multiple instances of zones list in the state. Not sure how that would work with modules, though.

Anyway, not in the scope of this PR, as we have similar solution for all generic data-sources at the moment.

kangasta (approved):
Could remove the dependency version downgrades, if those are not needed, but otherwise LGTM!

go.mod
villevsv-upcloud: Leftovers, removed now! :+1:

internal/service/cloud/zones_data_source.go
villevsv-upcloud: That sounds like a good idea.

PR #573, opened 2024-07-01
 refactor(firewall): migrate firewall_rules resource to plugin framework

villevsv-upcloud:
TODO:

- [ ] Use TFProto v6


internal/service/firewall/firewall.go
```diff
@@ -2,149 +2,187 @@ package firewall
...
+	resp.Schema = schema.Schema{
+		Description: "FirewallRuless can be used to connect multiple Private Networks. UpCloud Servers on any attached network can communicate directly with each other.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"server_id": schema.StringAttribute{
 				Required:    true,
-				ForceNew:    true,
+				Description: "The unique id of the server to be protected the firewall rules",
```
kangasta, line 77: I think I have used UUID in the other new Id descriptions, but this is a bit different since this refers Id of another resource  🤔

```suggestion
				Description: "The UUID of the server to be protected the firewall rules",
```

internal/service/firewall/firewall.go
```diff
@@ -2,149 +2,187 @@ package firewall
...
 			},
-			"firewall_rule": {
-				Type: schema.TypeList,
+			"firewall_rule": schema.ListNestedAttribute{
```
kangasta, line 79: Hmm, would `ListNestedBlock` work if we move this from `Attributes` to `Blocks` 🤔 Then this would not require change to protocol version 6 and configuration syntax would stay the same as well.

kangasta:
I cherry-picked the migrate commit on top of main to update this: closing this in favor of #771

PR #583, opened 2024-07-04
feat(kubernetes): update node group storage encryption validator

peknur:
Allow disabling cluster level storage encryption in node group.

### DIFF
--- a/internal/service/kubernetes/node_group.go
+++ b/internal/service/kubernetes/node_group.go
@@ -144,7 +144,17 @@ func ResourceNodeGroup() *schema.Resource {
 				Default:     true,
 				ForceNew:    true,
 			},
-			"storage_encryption": storageEncryptionSchema("Storage encryption strategy for the nodes in this group.", true),
+			"storage_encryption": {
+				Description: "Storage encryption strategy for the nodes in this group.",
+				Type:        schema.TypeString,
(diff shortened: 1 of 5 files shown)


internal/service/kubernetes/node_group.go
```diff
@@ -144,7 +144,17 @@ func ResourceNodeGroup() *schema.Resource {
...
 			},
-			"storage_encryption": storageEncryptionSchema("Storage encryption strategy for the nodes in this group.", true),
+			"storage_encryption": {
+				Description: "Storage encryption strategy for the nodes in this group.",
+				Type:        schema.TypeString,
+				Computed:    true,
+				Optional:    true,
+				ForceNew:    true,
+				ValidateDiagFunc: validation.ToDiagFunc(validation.StringInSlice([]string{
+					string(upcloud.StorageEncryptionDataAtReset),
```
kangasta, line 154: Should this be _data at rest_ instead of _data at reset_? Would have to be fixed in the Go SDK, though 🤔

kangasta (approved):

internal/service/kubernetes/node_group.go
peknur: Ah, good catch :smile: :+1: I'll create SDK PR to fix that.

PR #624, opened 2024-08-20
fix(kubernetes): do not set default for `storage_encryption`

kangasta:
In `upcloud_kubernetes_node_group:`, do not set default value for `storage_encryption` in the provider implementation. Instead, use value returned from the API when updating the node group state. This fixes data consistency error when creating node group without defining value for `storage_encryption` and does not try to replace node group when running apply after updating provider to `v5.10.0`.


upcloud/testdata/upcloud_kubernetes/kubernetes_labels_s2.tf
```diff
@@ -29,6 +29,7 @@ resource "upcloud_kubernetes_cluster" "main" {
   name                    = "${var.basename}cluster"
   network                 = upcloud_network.main.id
   zone                    = var.zone
+  storage_encryption      = "data-at-rest"
```
kangasta, line 32: Step 1 sets `storage_encryption`, so set it also here and in step 3 to avoid replacing the cluster during the test case.

internal/service/networkpeering/networkpeering.go
```diff
@@ -296,14 +296,20 @@ func waitForPeeringToLeaveProvisionedState(ctx context.Context, svc *service.Ser
...
+				utils.ErrorDiagnosticDetail(ctx.Err()),
+			)
 			return
 		default:
 			peering, err := svc.GetNetworkPeering(ctx, &request.GetNetworkPeeringRequest{
 				UUID: uuid,
 			})
 			if err != nil {
-				diags.AddError("Unable to get network peering details", ctx.Err().Error())
+				diags.AddError(
```
kangasta, line 309: Not related, but caused segfault in the acceptance tests triggered by this PR, so might as well fix here 🙈

peknur (approved):

PR #627, opened 2024-09-04
refactor(storage): migrate data source to plugin framework

kangasta:


internal/service/storage/storage_data_source.go
```diff
@@ -0,0 +1,232 @@
...
+				Optional:           true,
+			},
+			"name": schema.StringAttribute{
+				Description:        "Exact name of the storage (same as title). Deprecated, use `title` instead.",
+				DeprecationMessage: "Contains the same value as `title`. Use `title` instead.",
+				Optional:           true,
+			},
+			"most_recent": schema.BoolAttribute{
+				Description:        "If more than one result is returned, use the most recent storage. This is only useful with private storages. Public storages might give unpredictable results.",
+				DeprecationMessage: "Use exact title or UUID to limit the number of matching storages to one.",
```
villevsv-upcloud, line 122: Not a blocker, wondering if this could be re-phrased somehow as exact title might not help limiting the matches to 1, as it is not enforced to be unique.

villevsv-upcloud (approved):

internal/service/storage/storage_data_source.go
kangasta: Maybe like this?

```suggestion
				DeprecationMessage: "Use exact title or UUID to limit the number of matching storages. Note that if you have multiple storages with the same title, you should use UUID to select the storage.",
```

villevsv-upcloud: Nice

Issue #630, opened 2024-09-10
Support for importing compressed images via direct_upload 


james-monio:
I'm uploading some vm images to upcloud using the terraform provider to use as custom os templates, and I've found that uploading xz-compressed images does not work.

This is (approximately) what my terraform file looks like:

```
#storage import works, and is bootable when I turn it into a template
resource "upcloud_storage" "custom-os" {
  size  = 10
  tier  = "maxiops"
  title = "custom-os"
  zone  = "fi-hel1"

  import {
    source          = "direct_upload"
    source_location = "/tmp/custom-os.raw"
    source_hash     = filesha256("/tmp/custom-os.raw")
  }
}

# storage import works, but is not bootable
# not getting decompressed?
resource "upcloud_storage" "custom-os-compressed" {
  size  = 10
  tier  = "maxiops"
  title = "custom-os-compressed"
  zone  = "fi-hel1"

  import {
    source          = "direct_upload"
    source_location = "/tmp/custom-os.raw.xz"
    source_hash     = filesha256("/tmp/custom-os.raw.xz")
  }
}
```

After talking with some of the UpCloud team, it seems like the issue is likely that there's no support for adding a content type to the upload through the provider, so the imported file is always getting treated as a raw image.

kangasta:
Thank you for reporting this issue! We will take a look on how the content-type and content-encoding are set when creating the direct upload 👀

kangasta:
There is now a fix available in the [v5.11.0](https://github.com/UpCloudLtd/terraform-provider-upcloud/releases/tag/v5.11.0). The missing content-type was indeed the problem, and from v5.11.0 the content-type is set based on the file-extension of the file to be uploaded.

kangasta:
I'll close this as completed. Let's create a new issue if there is need to configure the header manually (instead of setting it based on file-extension).

PR #631, opened 2024-09-10
fix(storage): set `Content-Type` based on file extension

kangasta:
See #630 for the bug description.


uc-snyk (approved):
Looks good. We could add a mention about this new behavior in `source_location` description.

villevsv-upcloud (approved):
Looks good. We could add a mention about this new behavior in source_location description.

PR #652, opened 2024-10-07
refactor(ip): ip_addresses data source to plugin fw

villevsv-upcloud:


internal/service/ip/ip_addresses_data_source.go
```diff
@@ -0,0 +1,149 @@
...
+							Computed:            true,
+							MarkdownDescription: "A reverse DNS record entry",
+						},
+						"server": schema.StringAttribute{
+							Computed:            true,
+							MarkdownDescription: "The unique identifier for a server",
+						},
+						"zone": schema.StringAttribute{
+							Computed:            true,
+							MarkdownDescription: "Zone of address, required when assigning a detached floating IP address, e.g. `de-fra1`. You can list available zones with `upctl zone list`.",
```
kangasta, line 101: Read-only fields here, so could remove instructions to list available options:

```suggestion
							MarkdownDescription: "Zone of the IP address.",
```

internal/service/ip/ip_addresses_data_source.go
```diff
@@ -0,0 +1,149 @@
...
+
+	ipAddresses, err := d.client.GetIPAddresses(ctx)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Unable to read ip addresses",
+			utils.ErrorDiagnosticDetail(err),
+		)
+	}
+
+	data.ID = types.StringValue(time.Now().UTC().String())
```
kangasta, line 123: I guess the ID could also be something static for data sources that do not have any user defined fields, but maybe better to go through all of those at once 🤔

kangasta (approved):

internal/service/ip/ip_addresses_data_source.go
villevsv-upcloud: I would keep it as it is for now, but indeed go through them in a major release.

PR #658, opened 2024-10-08
feat(objsto): add bucket resource

kangasta:
- [x] Use released version of Go SDK, depends on https://github.com/UpCloudLtd/upcloud-go-api/pull/338
- [x] `go mod tidy`
- [x] Acceptance tests
- [x] See if  `updated_at` should be ignored by `ImportStateVerify`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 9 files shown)


back-2-95:
@kangasta what's your plan on this? This would allow us to remove aws provider to create buckets...

kangasta:
Sorry for the delay with this. The problem is that our own bucket API end-points are still quite limited, so this resource would only allow bucket creation and deletion at first. Thus, you would still need external resources to manage policies and other more advanced features.

villevsv-upcloud (approved):

PR #681, opened 2024-11-11
feat(server): state upgrader for index

villevsv-upcloud:


internal/service/server/server.go
```diff
@@ -440,6 +454,58 @@ func ResourceServer() *schema.Resource {
...
+	for _, n := range networking.Interfaces {
+		if n.Type != iface["type"].(string) {
+			continue
+		}
+
+		if n.MAC != iface["mac_address"].(string) {
+			continue
+		}
+
+		return n.Index, nil
```
kangasta, line 503: Is this same as:

```suggestion
		if n.Type == iface["type"].(string) && n.MAC == iface["mac_address"].(string) {
			return n.Index, nil
		}
```

No clear preference on which one would be better, though

PR #697, opened 2025-01-20
docs: update firewall_rule action description

MarcusC-UpCloud:
Add a description for possible action, Default action values either "Accept" or "Drop"

### DIFF
--- a/internal/service/firewall/firewall.go
+++ b/internal/service/firewall/firewall.go
@@ -57,7 +57,7 @@ func ResourceFirewallRules() *schema.Resource {
 						},
 						"action": {
 							Type:         schema.TypeString,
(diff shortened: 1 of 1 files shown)


villevsv-upcloud:
Thanks for the collaboration!

Few notes:

The title would need to include a conventional commit message prefix, in this case it would be `docs: `. 

Also, the docs themselves can't be modified directly as they are generated from the source code. This particular change would be done in https://github.com/UpCloudLtd/terraform-provider-upcloud/blob/main/internal/service/firewall/firewall.go#L60 .

MarcusC-UpCloud:
Hey!

The git commit has been changed to the correct one for the .go file

A new title was given to the PR in line with the standard naming convention

PR #719, opened 2025-02-27
chore: prepare release v5.20.1

kangasta:


kangasta:
I'll create the release tomorrow and will include #718 as well. Better to not merge #718 until before tomorrow in case the nightly workflow finds some changes 🙈

PR #735, opened 2025-03-14
ci: create releases from GitHub actions

kangasta:


.github/workflows/release.yml
kangasta, line 1: This is copied from our objsto provider, which was created with the TF boilerplate

PR #739, opened 2025-03-14
feat(storage backup): Implementation of upcloud_storage_backup

paketeserrano:
- Added new storage backup resource: upcloud_storage_backup
- Added TestAccUpCloudStorageBackup_basic test
- Example:
```shell
resource "upcloud_storage_backup" "backup"{
  title = "backup13"
  source_storage = "0185b32e-41fd-46e4-b32e-a9d9028eaec6"
}
``` 
Notes: The resource manages one backup. Updates only changes the title. Deleting the resource from the configuration will delete the managed backup

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 
 - Display descriptive error message when credentials are not configured.
+
+### Added
+
 - Added upcloud_storage_backup resource that creates a backup of a storage resource
 
 ## [5.20.3] - 2025-03-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
+
+- Added upcloud_storage_backup resource that creates a backup of a storage resource
+
 ## [5.21.0] - 2025-04-15
 
 ### Added
@@ -38,10 +42,6 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 
 - Display descriptive error message when credentials are not configured.
-
-### Added
-
-- Added upcloud_storage_backup resource that creates a backup of a storage resource
 
 ## [5.20.3] - 2025-03-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 
 - Display descriptive error message when credentials are not configured.
+- Added upcloud_storage_backup resource that creates a backup of a storage resource
 
 ## [5.20.3] - 2025-03-06
 
--- a/internal/service/storage/storage_backup.go
+++ b/internal/service/storage/storage_backup.go
@@ -8,7 +8,6 @@ import (
 	"github.com/UpCloudLtd/upcloud-go-api/v8/upcloud"
 	"github.com/UpCloudLtd/upcloud-go-api/v8/upcloud/request"
 	"github.com/UpCloudLtd/upcloud-go-api/v8/upcloud/service"
-	"github.com/hashicorp/terraform-plugin-framework/path"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
@@ -17,9 +16,8 @@ import (
 )
 
 var (
-	_ resource.Resource                = &storageBackupResource{}
-	_ resource.ResourceWithConfigure   = &storageBackupResource{}
-	_ resource.ResourceWithImportState = &storageBackupResource{}
+	_ resource.Resource              = &storageBackupResource{}
+	_ resource.ResourceWithConfigure = &storageBackupResource{}
 )
 
 type storageBackupResource struct {
@@ -120,7 +118,6 @@ func (r *storageBackupResource) Read(ctx context.Context, req resource.ReadReque
 		return
 	}
 
-	// TODO: Do we have to remove the resource or do we have to log an error and return?
 	if data.SourceStorage.ValueString() == "" {
 		resp.State.RemoveResource(ctx)
 		return
@@ -214,6 +211,7 @@ func (r *storageBackupResource) Delete(ctx context.Context, req resource.DeleteR
 		return
 	}
 
+	// Do we want to delete the snapshot from the system or only make TF think it is deleted?
 	deleteStorageRequest := &request.DeleteStorageRequest{
 		UUID: backupID,
 	}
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 - Display descriptive error message when credentials are not configured.
 
+### Added
```
kangasta, line 12: Might need to rebase this branch to get this to correct section as there are new releases

internal/service/storage/storage_backup.go
```diff
@@ -0,0 +1,227 @@
...
+	SourceStorage types.String `tfsdk:"source_storage"`
+	Title         types.String `tfsdk:"title"`
+	ID            types.String `tfsdk:"id"`
+	CreatedAt     types.String `tfsdk:"created_at"`
+}
+
+func (r *storageBackupResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages an on-demand storage backup.",
+		Attributes: map[string]schema.Attribute{
```
kangasta, line 50: We could use `Computed` attributes to provide values of the other storage fields (e.g. `size`, `type`, `zone`, ...) as well, but this can also be done in a later PR. See template resource for example.

Could also see if `storageCommonModel` would help with defining those additional fields 🤔

internal/service/storage/storage_backup.go
```diff
@@ -0,0 +1,227 @@
...
+
+func (r *storageBackupResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	var data storageBackupModel
+	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if data.SourceStorage.ValueString() == "" {
```
kangasta, line 121: Should this be `ID` instead of `SourceStorage`? I think the backup can be used also after the source storage has been removed 🤔 

```suggestion
	if data.ID.ValueString() == "" {
```

internal/service/storage/storage_backup.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func (r *storageBackupResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manages an on-demand storage backup.",
+		Attributes: map[string]schema.Attribute{
+			"source_storage": schema.StringAttribute{
+				Required:    true,
+				Description: "The UUID of the storage to back up.",
+			},
```
kangasta, line 54: Could use `RequiresReplace` plan modifier here (similarly than in template resource):

```suggestion
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.RequiresReplace(),
				},
			},
```

internal/service/storage/storage_backup.go
```diff
@@ -0,0 +1,227 @@
...
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Check if the source storage has changed
+	if data.SourceStorage.ValueString() != state.SourceStorage.ValueString() {
+		resp.Diagnostics.AddError("Source_storage cannot be changed", "Source_storage is immutable and cannot be changed.")
+		return
+	}
```
kangasta, line 168: This would not be needed if `RequiresReplace` plan modifier is defined for the field (see comment in schema)

upcloud/resource_upcloud_storage_test.go
```diff
@@ -550,3 +550,78 @@ func createTempImage() (string, *hash.Hash, error) {
...
+					resource.TestCheckResourceAttr(resourceName, "title", initialBackupTitle),
+				),
+			},
+			// Step 2: Update the Backup Title
+			{
+				Config: testUpcloudStorageBackupConfig(updatedBackupTitle),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(resourceName, "title", updatedBackupTitle),
+				),
+			},
```
kangasta, line 586: Could add step to verify import operation, something like:

```
{
	ResourceName:       resourceName,
	ImportState:        true,
	ImportStateVerify:  true,
},
```

kangasta (commented):
Could use `RequiresReplace` for `source_storage`, check field name in update function, and add test step to verify import, otherwise looks good to me.

CHANGELOG.md
paketeserrano: yep, I am expecting that. I will rebase and fix if needed

upcloud/resource_upcloud_storage_test.go
kangasta: This might also need `ImportStateVerifyIgnore: []string{"created_at"}`, if that field is not available in the GetStorage output for backups 🤔

internal/service/storage/storage_backup.go
paketeserrano: Like it

internal/service/storage/storage_backup.go
paketeserrano: It will be removed as I added the RequiresReplace for that field

internal/service/storage/storage_backup.go
paketeserrano: I implemented this change and I got this, which makes sense. I am not sure if it is usable from the point of view of terraform, like how these backups would be used later on. At least, we keep them just in case.
Screenshot from 2025-04-22 19-38-50

upcloud/resource_upcloud_storage_test.go
paketeserrano: Done

paketeserrano:
@kangasta  I implemented the changes you mentioned in your review. Check it out and let me know. I created a new ticket  to implement the computed properties and will be implemented in a different PR

internal/service/storage/storage_backup.go
```diff
@@ -0,0 +1,234 @@
...
+	})
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Storage did not reach online state",
+			utils.ErrorDiagnosticDetail(err),
+		)
+		return
+	}
+
+	// Do we want to delete the snapshot from the system or only make TF think it is deleted?
```
kangasta, line 213: I guess expected behavior would be to delete the backup. If we only remove it from Terraform, customer will continue to pay for the backup even though they think it was deleted 🤔 

```suggestion
```

upcloud/resource_upcloud_storage_test.go
```diff
@@ -550,3 +550,108 @@ func createTempImage() (string, *hash.Hash, error) {
...
+
+		// Step 2: Create a backup from the storage
+		resource "upcloud_storage_backup" "this" {
+			source_storage = upcloud_storage.test.id
+			title          = "%s"
+		}
+	`, backupTitle)
+}
+
+func TestAccUpCloudStorageBackup_import(t *testing.T) {
```
kangasta, line 629: This could also be a test step in the existing test, but no big difference in this case as the storage resources are relatively fast to create and backup 🤔

kangasta (approved):

PR #747, opened 2025-04-10
feat(server): implement hot resize

thevilledev:
Add support for resizing server plans without requiring a reboot using the new `hot_resize` attribute. When enabled, this feature attempts to change the server plan without stopping the server. If hot resize fails, the apply operation will fail with a clear error message.

Key changes:
- Add `hot_resize` attribute to the server resource schema
- Implement logic to detect when server restart is required
- Update server modification code to attempt hot resize when applicable
- Add comprehensive tests for various hot resize scenarios
- Improve error handling to provide clear guidance when hot resize fails

This feature provides users with a way to change server plans with zero downtime when the hot resize feature is available. If the hot resize operation fails, users can set `hot_resize = false` to use the standard approach that includes a server restart.

### References

- https://developers.upcloud.com/1.3/8-servers/#modify-server
- https://upcloud.com/docs/guides/scale-cloud-servers-hot-resize/

### DIFF
--- a/internal/service/server/server.go
+++ b/internal/service/server/server.go
@@ -94,6 +94,7 @@ type serverModel struct {
 	Login             types.Set    `tfsdk:"login"`
 	SimpleBackup      types.Set    `tfsdk:"simple_backup"`
 	BootOrder         types.String `tfsdk:"boot_order"`
+	HotResize         types.Bool   `tfsdk:"hot_resize"`
 }
 
 type networkInterfaceModel struct {
@@ -304,6 +305,12 @@ func (r *serverResource) getSchema(version int64) schema.Schema {
 				PlanModifiers: []planmodifier.String{
 					stringplanmodifier.UseStateForUnknown(),
 				},
+			},
+			"hot_resize": schema.BoolAttribute{
+				Description: "If set to true, allows changing the server plan without requiring a reboot. This enables hot resizing of the server. If hot resizing fails, the apply operation will fail.",
+				Optional:    true,
+				Computed:    true,
+				Default:     booldefault.StaticBool(false),
 			},
 		},
 		Blocks: map[string]schema.Block{
@@ -853,6 +860,11 @@ func setValues(ctx context.Context, data *serverModel, server *upcloud.ServerDet
 	data.Plan = types.StringValue(server.Plan)
 	data.BootOrder = types.StringValue(server.BootOrder)
 
+	// Set hot_resize to false by default if not set
+	if data.HotResize.IsNull() {
+		data.HotResize = types.BoolValue(false)
+	}
+
 	if !data.Tags.IsNull() {
 		data.Tags, diags = types.SetValueFrom(ctx, data.Tags.ElementType(ctx), server.Tags)
 		respDiagnostics.Append(diags...)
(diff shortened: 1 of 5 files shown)


upcloud/resource_upcloud_server_test.go
```diff
@@ -954,3 +957,142 @@ func TestUpcloudServer_createPreChecks(t *testing.T) {
...
+		}
+
+		output "server_ip" {
+			value = upcloud_server.hot_resize.network_interface[0].ip_address
+		}
+	`, planName, hotResize, keyDir, debianTemplateUUID, provisioner)
+}
+
+// generateSSHKey generates an SSH key pair in the given directory
+func generateSSHKey(t *testing.T, keyDir string) error {
```
kangasta, line 1032: Could also use [tls_private_key](https://registry.terraform.io/providers/hashicorp/tls/latest/docs/resources/private_key) resource to generate the key 🤔 Both approaches should work very similarly, though.

thevilledev: I think the downside of `tls_private_key` is that it's not a core Terraform resource. So it requires extra provider dependencies that will have to be initialised in `init()` alongside of the UpCloud provider. This requires a new Go dependency. Maybe opting in to the openssh CLI interface just might be better?

thevilledev: Actually maybe a pure Go implementation for key generation would be better. Maybe someone wants to run acceptance tests on Windows. :)

thevilledev: Done in https://github.com/UpCloudLtd/terraform-provider-upcloud/pull/747/commits/330cf5df839d7f31be77d33233661db76b9bb878

internal/service/server/server.go
```diff
@@ -1214,7 +1226,31 @@ func (r *serverResource) Update(ctx context.Context, req resource.UpdateRequest,
...
+		_, err := r.client.ModifyServer(ctx, hotResizeReq)
+		if err != nil {
+			// If hot resize fails, return an error
+			resp.Diagnostics.AddError(
+				"Hot resize failed",
+				fmt.Sprintf("Unable to hot resize server: %s. Hot resize requires the server to be in a state where it can be resized without a reboot. If hot resize is not possible, set hot_resize = false to use the standard approach with server reboot.", utils.ErrorDiagnosticDetail(err)),
+			)
+			return
+		}
+	} else if changeRequiresServerStop(state, plan) {
```
kangasta, line 1253: Can this `else if` cause problems if same plan contains hot-resize change and changes that require server restart 🤔

thevilledev: Yeah I agree, might become an issue. I think we could change the order? Pseudo code:

```
if needsServerStop {
   // ...
} else if planChanged && hotResizeEnabled {
   // ...
}
```

thevilledev: I built a test case around this. Creates a server with a plan + hot resizing, once applied it adds a utility network but also changes the plan. The test fails because:

```
        Error: Invalid server state

          with upcloud_server.mixed_changes,
          on terraform_plugin_test.tf line 12, in resource "upcloud_server" "mixed_changes":
          12:                                   resource "upcloud_server" "mixed_changes" {

        Server needs to be stopped to alter networks
```

I'll add a fix for this, thanks Toni

thevilledev: Fix in https://github.com/UpCloudLtd/terraform-provider-upcloud/pull/747/commits/fc40daa1372fab7717a0e80335ab60c12a45ecf6. Ran the tests locally and seems to be all fine, just takes 3-5 minutes to finish.

kangasta: Yeah, the reverse ordering should be safer 👌

kangasta (approved):

kangasta:
LGTM! Could add changelog entry as well 👍

thevilledev:
Thanks, tests passed so I'll merge it in!

PR #754, opened 2025-04-15
fix(server): config loading in Update

thevilledev:
The Update method was incorrectly attempting to load the Terraform configuration into the state variable instead of the config variable.

Consequently, further in the method this call has previously received an empty Host value:

```
	server, err := utils.VerifyServerStarted(ctx, request.StartServerRequest{UUID: uuid, Host: int(config.Host.ValueInt64())}, r.client)

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 
 - upcloud_server: add `hot_resize` attribute, allowing plan changes without server restarts when supported by the platform. See [UpCloud documentation](https://upcloud.com/docs/guides/scale-cloud-servers-hot-resize/) for more information.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 
 - upcloud_server: add `hot_resize` attribute, allowing plan changes without server restarts when supported by the platform. See [UpCloud documentation](https://upcloud.com/docs/guides/scale-cloud-servers-hot-resize/) for more information.
+- upcloud_server: Correctly load configuration in Update method, fixing an issue where the `Host` field was not populated, potentially causing resource update failures.
```
kangasta, line 9: These seem to be missing the type headings:

```suggestion

### Fixed

- upcloud_server: Correctly load configuration in Update method, fixing an issue where the `Host` field was not populated, potentially causing resource update failures.
```

kangasta (approved):
Good catch!

PR #761, opened 2025-04-24
feat(kubernetes): add cluster upgrade support

kangasta:


internal/service/kubernetes/kubernetes.go
```diff
@@ -28,6 +28,7 @@ const (
 	privateNodeGroupsDescription        = "Enable private node groups. Private node groups requires a network that is routed through NAT gateway."
 	stateDescription                    = "Operational state of the cluster."
 	versionDescription                  = "Kubernetes version ID, e.g. `1.31`. You can list available version IDs with `upctl kubernetes versions`."
```
villevsv-upcloud, line 30:
```suggestion
	versionDescription                  = "Kubernetes version ID, e.g. `1.31`. You can list available version IDs with `upctl kubernetes versions`.

    Note that when upgrading the cluster to a newer version, `upgrade_strategy` will be taken into account."
```

internal/service/kubernetes/kubernetes.go
```diff
@@ -28,6 +28,7 @@ const (
...
+	upgradeStrategyDescription          = "The upgrade strategy for the cluster when modifying the `version`."
```
villevsv-upcloud, line 31:
```suggestion
	upgradeStrategyDescription          = "The upgrade strategy for the cluster when modifying the `version`. If not set, upgrading the `version` will use the default strategy `manual`."
```

villevsv-upcloud (approved):

PR #762, opened 2025-04-29
feat(storage backup): Add computed values

paketeserrano:


internal/service/storage/storage_backup.go
```diff
@@ -75,6 +75,40 @@ func (r *storageBackupResource) Schema(_ context.Context, _ resource.SchemaReque
...
+			},
+			"type": schema.StringAttribute{
+				MarkdownDescription: typeDescription,
+				Computed:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"zone": schema.StringAttribute{
+				MarkdownDescription: "The zone the storage is in, e.g. `de-fra1`. You can list available zones with `upctl zone list`.",
```
kangasta, line 106: Could omit the list command from here (similarly as with templates) as zone is computed only 🤔 

```suggestion
				MarkdownDescription: "The zone the storage is in, e.g. `de-fra1`.",
```

kangasta (approved):

PR #763, opened 2025-04-30
feat(fix tests): fix acceptance test failures

paketeserrano:


.github/workflows/terraform.yml
```diff
@@ -30,8 +30,8 @@ jobs:
         run: make test
       - name: Run acceptance tests
         env:
-          UPCLOUD_USERNAME: ${{ secrets.UPCLOUD_USERNAME }}
-          UPCLOUD_PASSWORD: ${{ secrets.UPCLOUD_PASSWORD }}
+          UPCLOUD_USERNAME: ${{ secrets.UPCLOUD_USERNAME_ALT }}
```
kangasta, line 33: Should we just change the values of the existing secrets? I guess no need to store the old credentials in GitHub secrets anymore? 🤔

paketeserrano: Yes, my plan is to do that once I fix the tests errors we have so far and everything works

kangasta (approved):

PR #771, opened 2025-05-16
refactor(firewall): migrate firewall_rules resource to plugin framework

kangasta:


internal/service/firewall/firewall.go
```diff
@@ -29,127 +33,166 @@ The firewall is enabled on public and utility network interfaces.`
...
+	resp.TypeName = req.ProviderTypeName + "_firewall_rules"
+}
+
+// Configure adds the provider configured client to the resource.
+func (r *firewallRulesResource) Configure(_ context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
+	r.client, resp.Diagnostics = utils.GetClientFromProviderData(req.ProviderData)
+}
+
+type firewallRulesModel struct {
+	ID           types.String        `tfsdk:"id"`
```
kangasta, line 60: Check if ID or serverID could be removed as you can only have one firewall config per server: deprecate either of these and ensure they always have the same value.

paketeserrano (approved):
I left a comment, other than that it is working and looking good.

internal/service/firewall/firewall.go
```diff
@@ -29,127 +33,178 @@ The firewall is enabled on public and utility network interfaces.`
...
-						"action": {
-							Type:         schema.TypeString,
-							Description:  "Action to take if the rule conditions are met. Valid values `accept | drop`",
-							Required:     true,
-							ForceNew:     true,
-							ValidateFunc: validation.StringInSlice([]string{"accept", "drop"}, false),
+						"comment": schema.StringAttribute{
+							Description: "A comment for the rule.",
+							Optional:    true,
+							Computed:    true,
```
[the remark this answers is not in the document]
kangasta: Yes: without computed, if user does not specify value in their configuration the value will be null and Terraform plugin framework will raise an error because that does not equal empty string. With computed, the value will be undefined and we are allowed to modify the value.

I.e., the plugin framework is very strict about types and values in state matching the configuration. With the SDKv2 this was not an issue and it only logged warnings in the logs.

internal/service/firewall/firewall.go
villevsv-upcloud: I think that `server_id` makes sense UX-wise, so at least I don't mind the duplicate data.

villevsv-upcloud (approved):

internal/service/firewall/firewall.go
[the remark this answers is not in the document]
paketeserrano: cool. Thanks for the info

internal/service/firewall/firewall.go
lmmz: SDN private network firewall rules are coming in couple of months, then we might have multiple rulesets per server. But the actual API implementation is still a bit open question - original goal was to keep it similar to these public/utility rulesets, but there's also been idas to create more generic security policies which would contain multiple ACLs rules.

PR #773, opened 2025-05-26
feat(supabase example): add configuration files

paketeserrano:
- Add configuration files that will spin a supabase server
- Add variables files that users can configure to set up the server


paketeserrano:
@kangasta  I run make docs and the only diffs were that it formatted the new files I added. You can check the diffs in my latest commit. In spite of that, I can change the location of these files but I am not sure the best place. Is the templates folder a good place? I do not remember if this folder is used somewhere else too

paketeserrano:
Thanks @kangasta for the review. I am going to reject this PR because I decided to add this code to the PaaS repo I created to host our initiative to develop PaaS services and stacks deployments: https://github.com/UpCloudLtd/paasup/tree/main/supabase-terraform

PR #821, opened 2025-09-02
fix: use same validation for taint keys and labels

### LINKED ISSUE #820: Valid node group taint key not accepted due to strict regex validator
While setting up a node group for [CloudNativePG](<https://cloudnative-pg.io/documentation/current/>), I noticed that the UpCloud Pulumi provider does not accept the taint [cnpg suggests using when reserving nodes for PostgreSQL workloads](<https://cloudnative-pg.io/documentation/current/architecture/#reserving-nodes-for-postgresql-workloads>). It seems to stem from the Terraform provider and [the regex validation it's using](<https://github.com/UpCloudLtd/terraform-provider-upcloud/blob/df3202120a475a8140aa4ff0e2d02bf00bc837c4/internal/service/kubernetes/node_group.go#L392>) for taint keys.

Unless there's some compelling reason to use such a strict validation (which there very well may be), an easy fix for this would be using [the same validator used for K8s labels](<https://github.com/UpCloudLtd/terraform-provider-upcloud/blob/df3202120a475a8140aa4ff0e2d02bf00bc837c4/internal/utils/labels.go#L21C47-L21C64>).

okuuva:
Closes #820.


internal/service/kubernetes/node_group.go
```diff
@@ -389,7 +389,7 @@ func (r *kubernetesNodeGroupResource) Schema(_ context.Context, _ resource.Schem
 							MarkdownDescription: "Taint key.",
 							Required:            true,
 							Validators: []validator.String{
-								stringvalidator.RegexMatches(regexp.MustCompile("^[a-zA-Z0-9-]+$"), "needs to match regexp ^[a-zA-Z0-9-]+$"),
+								stringvalidator.RegexMatches(utils.ValidLabelKeyRegExp, utils.InvalidLabelKeyMessage),
```
villevsv-upcloud, line 392: Instead of using these from the `utils` package, it'd be a bit simpler to introduce the variable and the constant in this file, for example:

```
var taintKeyRegexpStr = regexp.MustCompile("^[ -^`-~]+[ -~]*$") // Printable ASCII characters: ' ' (Space), ..., '^', '_', '`', ..., `~`
```

villevsv-upcloud (changes requested):
Hi! Thanks for the contribution.

I am suggesting a small change to the variable usage, but otherwise the basic principle seems to work as intended. :+1:

internal/service/kubernetes/node_group.go
okuuva: Sure. My reasoning was to ensure the validation regex would be the same for taint keys and labels as apparently they follow the same syntax (though I couldn't find any definitive source for that). But keeping them separate also makes sense in case they aren't actually identical and require tweaking later on.

Will separate the two.

villevsv-upcloud (approved):

PR #824, opened 2025-09-05
feat(network): add dhcp_route_config

paketeserrano:


internal/service/network/network.go
```diff
@@ -58,13 +59,25 @@ type networkModel struct {
...
-	Family           types.String `tfsdk:"family"`
-	Gateway          types.String `tfsdk:"gateway"`
+	Address                 types.String                  `tfsdk:"address"`
+	DHCP                    types.Bool                    `tfsdk:"dhcp"`
+	DHCPDefaultRoute        types.Bool                    `tfsdk:"dhcp_default_route"`
+	DHCPDns                 types.Set                     `tfsdk:"dhcp_dns"`
+	DHCPRoutes              types.Set                     `tfsdk:"dhcp_routes"`
+	Family                  types.String                  `tfsdk:"family"`
+	Gateway                 types.String                  `tfsdk:"gateway"`
+	DHCPRoutesConfiguration *dhcpRoutesConfigurationModel `tfsdk:"dhcp_routes_configuration"`
```
kangasta, line 69: I think there were some problems if the model is used here directly instead of types.Object, but can't remember what those were 😅

I guess we can look at this later if needed and if the tests are passing with this definition 🤔

internal/service/network/network.go
```diff
@@ -179,6 +192,64 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+											Description: "Exclude routes by their source.",
+											Optional:    true,
+											Computed:    true,
+											ElementType: types.StringType,
+											Validators: []validator.Set{
+												setvalidator.SizeBetween(0, 2),
+												setvalidator.ValueStringsAre(
+													stringvalidator.OneOf(
+														"static-route",
+														"router-connected-networks",
```
kangasta, line 217: Could use the consts from Go SDK to define these

internal/service/network/network.go
```diff
@@ -179,6 +192,64 @@ func (r *networkResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+											Description: "Only include routes of these types.",
+											Optional:    true,
+											Computed:    true,
+											ElementType: types.StringType,
+											Validators: []validator.Set{
+												setvalidator.SizeBetween(0, 2),
+												setvalidator.ValueStringsAre(
+													stringvalidator.OneOf(
+														"user",
+														"service",
```
kangasta, line 232: Could use the consts from Go SDK to define these

kangasta (approved):
Overall, looks good to me! Added few minor comments that we can resolve later

internal/service/network/network.go
paketeserrano: I read a bit about this. It looks like types.Object allows to distinguish better between null, empty and unknown but it is less type safe. I am having a similar problem with the lists inside the object, not easy to differentiate between those states (although it is not the same). I would like to learn how to fix this problem at some point. We can tackle this later

kangasta: The error in test results, `[...] .ip_network[0].dhcp_routes_configuration: was absent, but now present [...]`, might be related to this. I guess we have to leave the block as null, if user has not defined it, as we get the error even though all fields inside the block are computed 🤔

upcloud/resource_upcloud_network_test.go
```diff
@@ -265,6 +265,86 @@ func TestAccUpCloudNetwork_labels(t *testing.T) {
...
+					}
+				`,
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr("upcloud_network.test", "ip_network.0.dhcp_routes_configuration.effective_routes_auto_population.enabled", "true"),
+					resource.TestCheckTypeSetElemAttr("upcloud_network.test", "ip_network.0.dhcp_routes_configuration.effective_routes_auto_population.filter_by_destination.*", "10.30.0.0/24"),
+					resource.TestCheckTypeSetElemAttr("upcloud_network.test", "ip_network.0.dhcp_routes_configuration.effective_routes_auto_population.filter_by_destination.*", "172.16.0.0/22"),
+					resource.TestCheckTypeSetElemAttr("upcloud_network.test", "ip_network.0.dhcp_routes_configuration.effective_routes_auto_population.exclude_by_source.*", "static-route"),
+					resource.TestCheckTypeSetElemAttr("upcloud_network.test", "ip_network.0.dhcp_routes_configuration.effective_routes_auto_population.filter_by_route_type.*", "service"),
+				),
+			},
```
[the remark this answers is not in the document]
paketeserrano: I fixed the Read so the import works as expected

upcloud/resource_upcloud_network_test.go
```diff
@@ -265,6 +265,86 @@ func TestAccUpCloudNetwork_labels(t *testing.T) {
...
+
+						ip_network {
+							address            = "10.20.0.0/24"
+							dhcp               = true
+							dhcp_default_route = true
+							family             = "IPv4"
+							gateway            = "10.20.0.1"
+
+							dhcp_routes_configuration {
+								effective_routes_auto_population {
```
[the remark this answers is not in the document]
paketeserrano: Yep, this is done

upcloud/resource_upcloud_network_test.go
```diff
@@ -265,6 +265,86 @@ func TestAccUpCloudNetwork_labels(t *testing.T) {
...
```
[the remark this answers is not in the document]
paketeserrano: Yep, this is done

internal/service/network/network.go
paketeserrano: We have implemented this using SingleNestedAttribute as we talked about.

PR #883, opened 2025-12-05
refactor(dbaas): migrate database resources to plugin framework

kangasta:


internal/service/database/database.go
```diff
@@ -12,427 +12,548 @@ import (
...
+				utils.ErrorDiagnosticDetail(err),
+			)
+			return diags
 		}
 
-		tflog.Info(ctx, "managed database updated", map[string]interface{}{"uuid": d.Id(), "name": d.Get("name")})
-	}
-
-	// Modify powered state if no PostgreSQL version update is requested
-	if d.HasChange("powered") && !d.HasChange("properties.0.version") {
```
kangasta: Removed this special case as I assume the DB system would throw some errors if trying to upgrade version right after start of shut-down right after version upgrade without waiting for stable state between these actions 🤔

internal/service/database/properties/schema.go
```diff
@@ -110,127 +121,200 @@ func stringSlice(val interface{}) ([]string, bool) {
...
 			return nil, err
 		}
 
-		s.Type = schema.TypeList
-		s.MaxItems = 1
-		s.Elem = &schema.Resource{Schema: nested}
-	default:
-		return nil, fmt.Errorf(`unknown property value type "%s"`, prop.Type)
-	}
+		s := schema.ListNestedBlock{
```
kangasta: Figured it would be convenient to use same schema type as before, but looking at the complexity of the value conversions, switching to `SingleNestedBlock` (here and in nested object properties) and adding a `UpgradeState` might have been more simple 🤔

PR #895, opened 2026-01-09
chore: prepare release 5.33.0

paketeserrano:


kangasta (approved):
Looks good to me! Could use resource name (upcloud_managed_object_storage_policy) instead of _object storage policy version_ in the related changelog entry, but not a blocker

PR #896, opened 2026-01-09
feat: separate unit tests from acceptance tests

paketeserrano:


villevsv-upcloud (approved):
LGTM. I would use test instead of feat in the title :)

PR #925, opened 2026-02-06
fix(objsto): error if delete fails, remove deleted user policy from state

kangasta:


paketeserrano (approved):
good catch on the waiting when the object was already deleted

PR #930, opened 2026-02-11
chore: define and document end-to-end tests and their purpose

kangasta:


upcloud/test_helpers.go
```diff
@@ -31,13 +30,6 @@ func init() {
...
 
 func TestAccPreCheck(t *testing.T) {
-	if v := os.Getenv("UPCLOUD_USERNAME"); v == "" {
-		t.Fatal("UPCLOUD_USERNAME must be set for acceptance tests")
-	}
-	if v := os.Getenv("UPCLOUD_PASSWORD"); v == "" {
-		t.Fatal("UPCLOUD_PASSWORD must be set for acceptance tests")
-	}
-
 	err := TestAccProvider.Configure(context.Background(), terraform.NewResourceConfigRaw(nil))
```
kangasta: The configure function should fail if no credentials available, so no need to check the env variables before that.

villevsv-upcloud (approved):

PR #967, opened 2026-03-25
chore(objsto): bucket v9 api

villevsv-upcloud:


internal/service/managedobjectstorage/bucket.go
```diff
@@ -99,43 +125,82 @@ func (r *managedObjectStorageBucketResource) Create(ctx context.Context, req res
...
-			break
+	apiResp, err := client.ListObjectStorageBucketMetricsWithResponse(ctx, svcUUID, nil)
+	if err != nil {
+		diags.AddError(
+			"Unable to read managed object storage buckets",
+			utils.ErrorDiagnosticDetail(err),
+		)
+		return nil, diags
+	}
+	if apiResp.StatusCode() == http.StatusNotFound {
```
paketeserrano, line 179: Not sure if we need these 2:
apiResp.StatusCode() == http.StatusNotFound 
apiResp.StatusCode() != http.StatusOK

I noticed that the caller to this function do not know if the service did not exists or if the bucket did not exists. For instance, if the service does not exists we are not removing the bucket form the state

paketeserrano: I will approve anyway since my comment is a potential improvement

paketeserrano (approved):

internal/service/managedobjectstorage/bucket.go
villevsv-upcloud: I'll keep this in for now

PR #969, opened 2026-03-26
fix: wait for detach to finish

paketeserrano:


upcloud/server/resource_upcloud_server_test.go
```diff
@@ -2,6 +2,7 @@ package servertests
 
 import (
 	"fmt"
+	"math/rand"
```
scop: Insignificant as this is in tests, but using math/rand/v2 these days is arguably better in general, ref https://go.dev/blog/randv2

(Aside, there's another use of the old math/rand in upcloud/storage/resource_upcloud_storage_test.go)

villevsv-upcloud (approved):
A changelog entry would be good

upcloud/server/resource_upcloud_server_test.go
paketeserrano: I updated them

paketeserrano:
I added the changelog updates

PR #975, opened 2026-03-27
feat(objsto): upcloud_managed_object_storage_static_site resource

villevsv-upcloud:


internal/service/managedobjectstorage/static_site.go
```diff
@@ -0,0 +1,488 @@
...
+				Optional:    true,
+				Computed:    true,
+			},
+		},
+		Blocks: map[string]schema.Block{
+			"error_page": schema.ListNestedBlock{
+				MarkdownDescription: "Custom error pages served when the storage backend returns an error status.",
+				NestedObject: schema.NestedBlockObject{
+					Attributes: map[string]schema.Attribute{
+						"status_code": schema.Int64Attribute{
```
paketeserrano, line 119: Here we are going to need a custom validator to make sure that this definition from the api stands:

"Either status_code or both status_range.start and status_range.end must be provided, but not both"

internal/service/managedobjectstorage/static_site.go
```diff
@@ -0,0 +1,488 @@
...
+		)
+		return
+	}
+
+	if created.StatusCode() != http.StatusCreated {
+		diagUnexpectedStatus(&resp.Diagnostics, "create", created.StatusCode(), created.Body)
+		return
+	}
+
+	var dest v9.ObjectStorage2CreateStaticWebsite201
```
paketeserrano, line 303: This unmarshal business is not done in other resources. For instance, in user.go we do this
```
	apiResp, err := r.client.CreateObjectStorageUserWithResponse(ctx, svcUUID, v9.CreateObjectStorageUserJSONRequestBody{
		Username: data.Username.ValueString(),
	})
	if err != nil {
		resp.Diagnostics.AddError(
			"Unable to create managed object storage user",
			utils.ErrorDiagnosticDetail(err),
		)
		return
	}
	if apiResp.JSON201 == nil {
		resp.Diagnostics.AddError(
			"Unable to create managed object storage user",
			utils.ErrorDiagnosticDetail(fmt.Errorf("unexpected response: %s", apiResp.HTTPResponse.Status)),
		)
		return
	}
```
We assume the json201 contains the response. If there are cases that is not true then we will have to review them. I am also aware that we should always check the statusCode.

villevsv-upcloud: You are right, this was to work around something. Now the response handling is as elsewhere.

internal/service/managedobjectstorage/static_site.go
villevsv-upcloud: Added validation

internal/service/managedobjectstorage/static_site.go
```diff
@@ -0,0 +1,494 @@
...
+		resp.Diagnostics.AddError(
+			"Unable to create managed object storage static site",
+			utils.ErrorDiagnosticDetail(err),
+		)
+		return
+	}
+
+	if created.JSON201 == nil {
+		resp.Diagnostics.AddError(
+			"Unable to create managed object storage user",
```
paketeserrano: typo:  not creating a obj sto user

villevsv-upcloud: Fixed :+1:

internal/service/managedobjectstorage/custom_domain.go
```diff
@@ -227,33 +266,42 @@ func (r *managedObjectStorageCustomDomainResource) Update(ctx context.Context, r
...
+	apiResp, err := r.client.DeleteObjectStorageCustomDomainWithResponse(
+		ctx,
+		parsedUUID,
+		domainName,
+	)
 	if err != nil {
 		resp.Diagnostics.AddError(
 			"Unable to delete managed object storage custom domain",
 			utils.ErrorDiagnosticDetail(err),
 		)
```
paketeserrano, line 265: Missing a return afer adding the error to diagnostics, otherwise it keeps going and we most likely will have a panic

internal/service/managedobjectstorage/custom_domain.go
paketeserrano, line 132: We need to read the json201 from the response and copy it over to data and persist it in the state

paketeserrano, line 235: data.DomainName should be customDomain.DomainName (whatever is in the response) or move this line after the following if

internal/service/managedobjectstorage/custom_domain.go
villevsv-upcloud: Added the return

internal/service/managedobjectstorage/custom_domain.go
villevsv-upcloud: Totally missed the whole basic logic, thanks for the thorough review :+1:

villevsv-upcloud: Now respecting the API response value for domain name when setting id :+1:

internal/service/managedobjectstorage/static_site.go
```diff
@@ -0,0 +1,494 @@
...
+	created, err := r.client.CreateStaticWebsiteWithResponse(ctx, svcUUID, body)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Unable to create managed object storage static site",
+			utils.ErrorDiagnosticDetail(err),
+		)
+		return
+	}
+
+	if created.JSON201 == nil {
```
paketeserrano, line 314: We should check first the status code:
if created.StatusCode() != http.StatusCreated {
    diagUnexpectedStatus(&resp.Diagnostics, "create", created.StatusCode(), created.Body)
    return
}
We can leave the defensive 'if' after

internal/service/managedobjectstorage/static_site.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+
+	var serviceUUID, domainName string
+	resp.Diagnostics.Append(utils.UnmarshalIDDiag(data.ID.ValueString(), &serviceUUID, &domainName)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	data.ServiceUUID = types.StringValue(serviceUUID)
```
paketeserrano, line 348: This is redundant as data struct already contain the serviceUUID in it.

kangasta: I think we need this for import. Then the state only contains value for id 🤔

internal/service/managedobjectstorage/static_site.go
```diff
@@ -0,0 +1,501 @@
...
+			"domain_name": schema.StringAttribute{
+				MarkdownDescription: "A custom domain in `static-website` mode attached to the service.",
+				Optional:            true,
+				Computed:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplaceIfConfigured(),
+				},
+			},
+			"id": schema.StringAttribute{
+				Description: "ID of the custom domain. ID is in {object storage UUID}/{domain name} format.",
```
paketeserrano: This should not say custom domain since it is for the static site

internal/service/managedobjectstorage/static_site.go
```diff
@@ -0,0 +1,501 @@
...
+	}
+
+	body := v9.CreateObjectStorageStaticWebsiteJSONRequestBody{
+		BucketName: data.BucketName.ValueString(),
+	}
+
+	if !data.DomainName.IsNull() && !data.DomainName.IsUnknown() && data.DomainName.ValueString() != "" {
+		body.DomainName = data.DomainName.ValueStringPointer()
+	}
+	if !data.BucketPrefix.IsNull() && !data.BucketPrefix.IsUnknown() {
```
paketeserrano, line 285: Something to keep in mind for the future, these guards checking nil and unknown are not really needed when using the ValueXXXXXPointer because we get the same behavior. It only makes sense the check for empty string since the zero value is "" .

kangasta: Yes, would be good to list some best practices in the DEVELOPING.md file 🤔

kangasta: ValueXXXXXPointer might be a bit problematic, as on unknown value it returns pointer to an empty value. I think we would need nil in many cases 🤔

kangasta (approved):

PR #985, opened 2026-04-21
chore: v9 objsto standardize response handling

paketeserrano:


villevsv-upcloud:
I think this is not needed as the SDK should do the unmarshaling. A previous of using this pattern was to due SDK not parsing the response properly because of wrong response header.

paketeserrano:
@villevsv-upcloud let's talk about this one on Monday. I know, as you said, that the json fields in the response were only populated when the content type was properly set. Did something change in the generated code?

Issue #994, opened 2026-05-05
Feature Request: Add support for encryption at rest in upcloud_file_storage


bramhoven:
UpCloud recently introduced the ability to encrypt data at rest for FileStorage via the Cloud Manager and API. However, this argument is currently missing from the upcloud_storage resource in the Terraform provider.
According to the UpCloud documentation, encryption **must be enabled at creation** (cannot be toggled for existing storage).
I would like to be able to manage this setting directly via Terraform.

### Proposed Configuration:
Adding an encrypted (boolean) argument to the resource like is done for the existing `upcloud_storage` resource:
```
resource "upcloud_storage" "encrypted_disk" {
  size  = 10
  tier  = "maxiops"
  title = "encrypted-storage-example"
  zone  = "fi-hel1"

  # New argument
  encrypt = true
}
```

### References:
•	**UpCloud Documentation:** [File Storage encryption at rest](https://upcloud.com/docs/products/file-storage/encryption-at-rest/)
•	**API Detail:** The encrypted parameter is available in the POST /1.3/file-storage endpoint, under the field `encrypted`

kangasta:
@bramhoven I noticed you have already worked on this in your fork, so I assigned this to you. Thank you for the contributions and raising the issue!

bramhoven:
@kangasta no worries! Just wanted this feature to be implemented before file storage was available in more regions :)

kangasta:
Soon released in v5.38.0 🚢

PR #1002, opened 2026-05-20
feat(file-storage): add encrypt field to upcloud_file_storage resource

bramhoven:
Exposes the UpCloud File Storage encryption-at-rest setting as an optional computed encrypt attribute. The field uses RequiresReplace since encryption can only be set at creation time and is immutable thereafter.
Depends on upcloud-go-api adding Encrypted to FileStorage and CreateFileStorageRequest. Resolves issue #994.

I have been able to test the working of it with the provider build locally.
Let me know if there is anything missing!

### DIFF
--- a/examples/resources/upcloud_file_storage/resource.tf
+++ b/examples/resources/upcloud_file_storage/resource.tf
@@ -16,6 +16,9 @@ resource "upcloud_file_storage" "example" {
   zone              = "fi-hel2"
   configured_status = "stopped"
 
+  # Encryption is set at creation and cannot be changed.
+  encrypt           = true
+
   labels = {
     environment = "staging"
     customer    = "example-customer"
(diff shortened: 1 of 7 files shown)


examples/resources/upcloud_file_storage/resource.tf
```diff
@@ -29,6 +29,15 @@ resource "upcloud_file_storage" "example" {
   }
 }
 
+# Encrypted file storage — encryption is set at creation and cannot be changed.
+resource "upcloud_file_storage" "encrypted" {
+  name              = "example-encrypted-file-storage"
+  size              = 250
+  zone              = "fi-hel2"
+  configured_status = "stopped"
+  encrypt           = true
```
kangasta: Let's add this to existing example to keep this shorter 🤔

internal/service/filestorage/file_storage.go
```diff
@@ -218,6 +230,7 @@ func (r *fileStorageResource) Create(ctx context.Context, req resource.CreateReq
 		SizeGiB:          int(data.Size.ValueInt64()),
 		Zone:             data.Zone.ValueString(),
 		ConfiguredStatus: upcloud.FileStorageConfiguredStatus(data.ConfiguredStatus.ValueString()),
+		Encrypted:        !data.Encrypt.IsNull() && !data.Encrypt.IsUnknown() && data.Encrypt.ValueBool(),
```
kangasta: `ValueBool` returns false on null or undefined, so I think the null and undefined checks can be skipped here 🤔 

```suggestion
		Encrypted:        data.Encrypt.ValueBool(),
```

kangasta (approved):
Overall looks good to me, thank you for the contribution! I did few small modifications and will merge this in a moment ⌛

bramhoven:
@kangasta alright perfect!

PR #1012, opened 2026-05-26
feat: firewall ruleset

paketeserrano:


internal/service/firewallruleset/rule.go
```diff
@@ -0,0 +1,556 @@
...
+	if !plan.Protocol.IsNull() && plan.Protocol.ValueString() != "" {
+		body.Protocol = plan.Protocol.ValueStringPointer()
+	}
+	if !plan.Enabled.IsNull() {
+		body.Enabled = plan.Enabled.ValueBoolPointer()
+	}
+	if !plan.Comment.IsNull() && plan.Comment.ValueString() != "" {
+		body.Comment = plan.Comment.ValueStringPointer()
+	}
+	if !plan.Position.IsNull() && plan.Position.ValueInt64() != 0 {
```
villevsv-upcloud, line 462: Should these be sent if the provided value is `0`? I think the spec allows a minimum value of `0` for position, icmp_type, source_port_start, source_port_end, destination_port_start and destination_port_end

paketeserrano: @villevsv-upcloud This is a good catch because the icmp_type can have 0 value according to the specs. The others have a range from 1 to xxx. I actually fixed the icmp in the update but I also added validators in the schema so there is better messages to the user and the int fields comply with the min and max values defined in the specs

internal/service/firewallruleset/ruleset.go
```diff
@@ -0,0 +1,404 @@
...
+
+	apiLabels := make([]v9.FirewallRulesetCreateLabel, 0, len(labelMap))
+	for _, k := range keys {
+		apiLabels = append(apiLabels, v9.FirewallRulesetCreateLabel{Key: k, Value: labelMap[k]})
+	}
+
+	return &apiLabels, nil
+}
+
+func setRulesetState(ctx context.Context, state *firewallRulesetModel, api *v9.FirewallRulesetDetailResponse) error {
```
villevsv-upcloud:
```suggestion
func setRulesetValues(ctx context.Context, state *firewallRulesetModel, api *v9.FirewallRulesetDetailResponse) error {
```

to match existing resources

villevsv-upcloud (approved):

internal/service/firewallruleset/ruleset.go
paketeserrano: done!

PR #1018, opened 2026-06-17
feat: firewall ruleset

paketeserrano:


go.sum
```diff
@@ -9,14 +9,14 @@ github.com/ProtonMail/go-crypto v1.1.6/go.mod h1:rA3QumHc/FZ8pAHreoekgiAbzpNsfQA
...
 github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260530122311-05187940889f/go.mod h1:Sy9fWRwKJnUZlN9LwnEwCOG1pgkktR303kD88V38D6E=
-github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260610080907-e5712c05723b h1:v07CE01bmlj3Xd61LZMwUK8UkPbqpaIDWVP6W6VvX3w=
-github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260610080907-e5712c05723b/go.mod h1:Sy9fWRwKJnUZlN9LwnEwCOG1pgkktR303kD88V38D6E=
-github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260611130528-1c6b6241a4c2 h1:FpmvNC13RG0Teev8mVhLOCmQ7qyDrH3rulbgayV/HZY=
-github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260611130528-1c6b6241a4c2/go.mod h1:Sy9fWRwKJnUZlN9LwnEwCOG1pgkktR303kD88V38D6E=
+github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260616113917-d8288b106ee6 h1:sX4sLdNKY+uauxDt8oygp27kwjXOkC0au8pBWWaTou4=
+github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260616113917-d8288b106ee6/go.mod h1:Sy9fWRwKJnUZlN9LwnEwCOG1pgkktR303kD88V38D6E=
+github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260617185413-91e273556c74 h1:WgSdTibC8FaoTDxzN+UdPZGZN9i+4Hp75Do37IXXioo=
+github.com/UpCloudLtd/upcloud-go-api/v9 v9.0.0-20260617185413-91e273556c74/go.mod h1:Sy9fWRwKJnUZlN9LwnEwCOG1pgkktR303kD88V38D6E=
```
kangasta: Could run `go mod tidy` to clean these up 🧹

upcloud/firewall/resource_upcloud_firewall_ruleset_test.go
kangasta, line 49: Could also check the resource action here with [plan checks](https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/plan-checks/resource#expectresourceaction-plan-check). E.g., check that action is `plancheck.ResourceActionUpdate` here 🤔

subcategories.json
kangasta: The rule resource can be removed from here, right? 🤔

internal/service/firewallruleset/ruleset.go
```diff
@@ -56,7 +110,7 @@ func (r *firewallRulesetResource) Configure(_ context.Context, req resource.Conf
 
 func (r *firewallRulesetResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
 	resp.Schema = schema.Schema{
-		Description: "This resource represents an UpCloud SDN firewall ruleset.",
+		Description: "This resource represents an UpCloud SDN firewall ruleset. Rules are managed as an ordered list; their position in the API is determined by their order in this list.",
```
kangasta: Could highlight the name of the field where the list is here, e.g.:

```suggestion
		Description: "This resource represents an UpCloud SDN firewall ruleset. Rules are managed as an ordered list; their position in the API is determined by their order in the `rules` list.",
```

internal/service/firewallruleset/ruleset.go
```diff
@@ -115,13 +162,352 @@ func (r *firewallRulesetResource) Schema(_ context.Context, _ resource.SchemaReq
...
+							Optional:    true,
+							Computed:    true,
+						},
+						"comment": schema.StringAttribute{
+							Description: "Rule comment.",
+							Optional:    true,
+							Computed:    true,
+						},
+						"position": schema.Int64Attribute{
+							Description: "Rule position (1-based). Computed from the rule's index in the list; cannot be configured.",
```
kangasta: In the docs this would be under _read-only_ title, so maybe no need for the clarification 🤔 

```suggestion
							Description: "Rule position (1-based). Computed from the rule's index in the list.",
```

internal/service/firewallruleset/ruleset.go
```diff
@@ -115,13 +162,352 @@ func (r *firewallRulesetResource) Schema(_ context.Context, _ resource.SchemaReq
...
+	r := ruleBlockModel{
+		Action:    types.StringValue(interfaceString(api.Action)),
+		Direction: types.StringValue(interfaceString(api.Direction)),
+		Family:    types.StringValue(interfaceString(api.Family)),
+	}
+	if api.Protocol == nil {
+		r.Protocol = types.StringNull()
+	} else {
+		r.Protocol = types.StringValue(interfaceString(api.Protocol))
+	}
```
kangasta, line 361: I think the pointer version of `types.*Value` might work for these. For example [StringPointerValue](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-framework/types#StringPointerValue) here 🤔 


```suggestion
    r.Protocol = types.StringPointerValue(api.Protocol)
```

internal/service/firewallruleset/ruleset.go
kangasta: Could also try here if the `types.*PointerValue` functions would work, to avoid having to check for nil value in each assignment 🤔

kangasta (commented):
We could try if we can use the pointer helpers from plugin framework when setting values, otherwise looks good to me!

go.sum
paketeserrano: done

subcategories.json
paketeserrano: This one was removed

internal/service/firewallruleset/ruleset.go
paketeserrano: applied

internal/service/firewallruleset/ruleset.go
paketeserrano: It did not work for protocol field but for some others

upcloud/firewall/resource_upcloud_firewall_ruleset_test.go
paketeserrano: done

PR #1020, opened 2026-06-23
feat: configure logging and user-agent for SDK v9 requests

kangasta:


upcloud/provider.go
```diff
@@ -120,6 +121,13 @@ func withInt64Default(val types.Int64, def int64) int64 {
 	return val.ValueInt64()
 }
 
+func withUserAgent(ua string) v9.ClientOption {
```
paketeserrano, line 124: ua param is not used. Is this intended?

paketeserrano:
Can we add some tests?

upcloud/provider.go
kangasta: Good catch! It should be used, to support overriding it (e.g. from Pulumi provider)

paketeserrano (approved):

PR #1023, opened 2026-07-03
add upcloud_server data source

dmechas:
- Adds a new `upcloud_server` data source that looks up an existing UpCloud server by id (UUID) and exposes its details
- Registers the new data source in the provider (upcloud/provider.go).

Notes:
- Some resource-only attributes (hot_resize, user_data, login, template) are intentionally not exposed by the data source. The UpCloud API doesn't return them on a server-details read
- Docs is not added


internal/service/server/server_data_source.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	details, err := d.client.GetServerDetails(ctx, &request.GetServerDetailsRequest{
+		UUID: data.ID.ValueString(),
+	})
+	if err != nil {
+		if utils.IsNotFoundError(err) {
+			resp.State.RemoveResource(ctx)
+			return
+		}
```
villevsv-upcloud: This should be removed altogether or RemoveResource should be replaced with AddError like below to be more straightforward and avoid missing state.

internal/service/server/server_data_source.go
```diff
@@ -0,0 +1,268 @@
...
+	Title             types.String `tfsdk:"title"`
+	VideoModel        types.String `tfsdk:"video_model"`
+	Zone              types.String `tfsdk:"zone"`
+}
+
+func (d *serverDataSource) Schema(_ context.Context, _ datasource.SchemaRequest, resp *datasource.SchemaResponse) {
+	additionalIPAddressAttrs := map[string]schema.Attribute{
+		"ip_address": schema.StringAttribute{
+			Computed:    true,
+			Description: "An additional IP address for this interface.",
```
villevsv-upcloud: (Applies to all the fields:) It would be better to use `MarkdownDescription` for consistency with other resources and data sources.

villevsv-upcloud (dismissed):
Hi! Thank you for your contribution and my apologies on the delay from our side.

I have added a couple of comments, but in general this looks useful and good to be merged. 👍 

If possible, please add the changelog entry and we can merge this after the requested changes and a passing pipeline.

Changelog entry could be something like:

```
## [Unreleased]

### Added

- upcloud_server (data source): data source for looking up an existing server by id
```

internal/service/server/server_data_source.go
```diff
@@ -0,0 +1,268 @@
...
+
+func (d *serverDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_server"
+}
+
+func (d *serverDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	d.client, resp.Diagnostics = utils.GetClientFromProviderData(req.ProviderData)
+}
+
+type serverDataSourceModel struct {
```
kangasta, line 37: If possible, would be also nice to have a model for the fields shared by resource and data-source and method for setting these common values. We currently have something like this in storage resources: [internal/service/storage/common.go](https://github.com/UpCloudLtd/terraform-provider-upcloud/blob/main/internal/service/storage/common.go)

dmechas:
thanks @kangasta and @villevsv-upcloud. please take a look

internal/service/server/server_data_source.go
dmechas: updated

internal/service/server/server_data_source.go
dmechas: updated

internal/service/server/server_data_source.go
dmechas: updated

kangasta:
Thank you for the updates, looks good to me now! 🙌

I'll do some testing in #1041. I'll then cherry-pick the new commit here, update the docs, and will then merge this one.

kangasta (approved):

Issue #1032, opened 2026-07-16
upcloud_gateway: API requires plan on creation, but schema marks it optional and the docs/example omit it


AbelMeijberg:
Creating a gateway without `plan` fails at apply time, including with the exact code from the docs' Example Usage:

  ```
  Error: Validation error. (status=400, invalid_params_(root)='Missing properties: 'plan'.')
  ```

  The schema marks `plan` as Optional ("Gateway pricing plan."), the example omits it, and the valid values aren't documented anywhere in the provider. I had to query `GET /1.3/gateway/plans`: `essentials`, `development`, `standard`, `production`, `advanced`. (Side note: the
  [product docs](https://upcloud.com/docs/products/nat-gateway/configurations/) list "Developer" instead of `development` and don't mention
  `essentials`.)

  **Expected:** the provider defaults or requires `plan` so this fails at plan time instead of mid-apply, or at minimum the docs say the API
  requires it and list the values.

  **Versions:** provider 5.41.0, OpenTofu 1.12.4 (tool-agnostic, the request is built by the provider).

  I'll submit a PR for the attribute description + example. Required-vs-default in the schema is a design decision I'll leave to you, happy
  to do a follow-up PR either way.

kangasta:
Thank you for the PR and reporting this! 🙌  Docs updates and default value now included in [v5.42.0](https://github.com/UpCloudLtd/terraform-provider-upcloud/releases/tag/v5.42.0) 🚢

PR #1033, opened 2026-07-16
fix(gateway): document that plan is required and list valid values

AbelMeijberg:
Relates to #1032.

 The UpCloud API rejects gateway creation without `plan`, but the schema description and the Example Usage don't mention it. This PR:

  - adds the valid plan values and the API requirement to the `plan` attribute description
  - adds `plan` to the example so it works when copy-pasted
  - regenerates the docs (`make docs`)

  Deliberately does not change the schema (required vs. default is left to #1032).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ See updating [Changelog example here](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


kangasta:
Thank you for the PR! 🙌 We have usually included the upctl command (instead of list of valid values) in similar cases, so we could follow the same pattern here 🤔 

I also added default value for plan for better backwards compatibility.

kangasta (approved):

Issue #961, opened 2026-03-19
Add support for labels to the `upcloud_loadbalancer_dynamic_certificate_bundle` resource


jcborras:
Little else to add besides the subject. My guess is the this limitation appears in many other resources as well.
I'm not discovering anything here, labels are supported by the API: https://developers.upcloud.com/1.3/17-managed-loadbalancer/#create-dynamic-certificate-bundle
Thanks for your time.

jcborras:
Default labels (ie. https://developer.hashicorp.com/terraform/tutorials/aws/aws-default-tags) would be the cherry at the top.
T, thanks for the comment on the issue.

kangasta:
I created a new issue to track the default labels feature request: #1056

PR #1044, opened 2026-07-30
feat(loadbalancer): add labels field to certificate bundle resources

kangasta:
Implements #961 

Also migrates the resources from v8 SDK to v9 SDK.


internal/service/loadbalancer/manual_certificate_bundle.go
```diff
@@ -234,42 +257,123 @@ func (r *manualCertificateBundleResource) Read(ctx context.Context, req resource
...
+			fmt.Sprintf("Unexpected API status code %d: %s", apiResp.StatusCode(), string(apiResp.Body)),
+		)
+		return
+	}
+
+	resp.Diagnostics.Append(setManualCertificateBundleValues(ctx, &data, apiResp.JSON200)...)
 	resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
 }
 
+func buildModifyReq(ctx context.Context, data manualCertificateBundleModel) (v9.ModifyLoadBalancerCertificateBundleJSONRequestBody, diag.Diagnostics) {
```
kangasta, line 287: This is quite hacky now, maybe more manual approach (like in [SDK v8](https://github.com/UpCloudLtd/upcloud-go-api/blob/main/upcloud/request/load_balancer.go#L732)) would be cleaner 🤔 Hopefully temporary hack as we should fix the type in OpenAPI spec or generated code, though.

villevsv-upcloud (approved):

